PR #31, opened 2021-12-24
Transition to lofty

Serial-ATA:
Signed-off-by: Serial <69764315+Serial-ATA@users.noreply.github.com>

Sorry for the big PR :)

I wasn't sure if you wanted to still use `GStreamer::duration`, so I left it in place. There were also some unused dependencies that I commented out. I've tested this out with a few songs and all seems to be working.


tramhao:
Thanks so much! My original plan is to migrate little by little, but there is no better PR than the original author so thanks for the great help and I'll merge it.

tramhao:
gstreamer duration is used before for correcting mp3 duration when the actual duration is different from tag info. For some file formats, the duration is not needed. How about lofty? I don't know if the duration of lofty for mp3 is from tag or from framecount/bitrate. Thanks.

PR #43, opened 2022-02-22
Update README

Serial-ATA:
The README was out of date, still treating GStreamer as the only backend in the `Installation` section. It should be clearer now :)

I also put the two screenshots into a table so they aren't as big.


tramhao:
Thanks so much for your help!

PR #49, opened 2022-03-08
Fix the placement of licenses

Serial-ATA:
Signed-off-by: Serial <69764315+Serial-ATA@users.noreply.github.com>

You had some licenses between imports, or missed some newlines :)

### DIFF
--- a/src/ui/components/key_editor/ke_input.rs
+++ b/src/ui/components/key_editor/ke_input.rs
@@ -1,9 +1,7 @@
 //! # Popups
 //!
-//!
 //! Popups components
 
-use super::{KeyBind, Keys};
 /**
  * MIT License
  *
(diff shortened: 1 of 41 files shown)


src/ui/components/key_editor/ke_input.rs
```diff
@@ -1,9 +1,7 @@
 //! # Popups
 //!
-//!
 //! Popups components
```
Serial-ATA, line 3: Can the doc comments like these be removed? I don't think they add much value, and they prevent the licenses from being at the top of the file.

Serial-ATA:
Also, unrelated to these changes, there were a lot of commented out imports. Should they be removed? I noticed a few were noted to be for debugging, but most can probably be taken out.

src/ui/components/key_editor/ke_input.rs
tramhao: Yes, sure.

tramhao:
Yes, thanks for your help:)

PR #64, opened 2022-04-13
Fix project name in README

Serial-ATA:
The contribution section had a different project's name in it :)


tramhao:
Thanks so much!

PR #84, opened 2022-07-07
Note the config file path for macOS

lxcode:
Just noting that the config.toml is in the Application Support dir on macOS.


tramhao:
Thanks for your help. As I'm not using macOS, if you meet any problem, don't hesitate to open an issue.

Issue #130, opened 2023-07-08
Implement a built-in web service to allow user to send request to control the termusic or get infromation of the player


pdckxd:
## Background
I am running termusic on a Raspberry Pi. I need to control the termusic from different devices such as macbook, windows laptop, cell phone and esp32 device. I understand dbus is an option but looks like it only supports limited player actions. For me, I am lazy to spend time on learning it :). I prefer that I can control the player via web api. Built-in web service is just a alternative for people who are familiar with web api. It's fun for me to write a bunch of code for this open source project and actually resolve my problem (maybe it can also help other people). Hopefully it can be merged finally. Thanks.

I've already implemented complete web service for termusic. Please assign the ticket to me. I will create PR soon.

tramhao:
Thanks so much for this big PR and it's very inspiring!
Currently I'm working on a similar solution in dev branch which use tonic as a grpc server. As the code is very different from v0.7.10, would you please check if this is a better start point for your request? commit 8dda is a working build.

pdckxd:
> Thanks so much for this big PR and it's very inspiring! Currently I'm working on a similar solution in dev branch which use tonic as a grpc server. As the code is very different from v0.7.10, would you please check if this is a better start point for your request? commit 8dda is a working build.

I looked dev branch. Seems your idea is to create dedicated service binary to handle remote request such as grpc, web api etc. It does make sense to use this pattern to avoid impacting termusic itself. I will take time to migrate my code based on this strategy.

tramhao:
Thanks so much! My idea is to split the server and client into separate workspace, thus the server can run without interface, client can communicate with server by grpc.

tramhao:
I wrap up the code in dev branch and publish it as v0.7.11. It's great to have the possibility to control it from other machines. I'll not modify too much on it except bug fixes thus feel free to change anything.

tramhao:
By the way, I just checked the cargo.lock file, and found tonic package use axum internally. I'm not familiar with axum, I saw your code is using axum. What is the relationship of tonic and axum? Which will be more convenient to use? Thanks.

pdckxd:
Looks like axum can use some mechanism to host both http & grpc services in same application.
See: https://github.com/tokio-rs/axum/blob/bb22802bfc3733ba2040de6ec0cfb8d3fd186fac/examples/rest-grpc-multiplex/src/main.rs

pdckxd:
[...]

PR #131, opened 2023-07-08
Implement a built-in web service to allow user to send request to con…

pdckxd:
…trol the termusic or get infromation of the player #130.

Detail instruction of web service can be found from HowtoUseBuiltinWebService.md


pdckxd:
Will refactor and move to dev branch based on tramhao's suggestion

Issue #150, opened 2023-09-22
`playing` icon in front of currently playing track and currently playing track in status bar


GaelicGabe:
I am missing a feature where the currently playing track has some form of indication that it's playing.
I would therefore like to make the following changes:

* Add customizable icon in front of the track in the playlist menu (exactly where I am still unsure of)
* Add track name in the status menu

Since I would also like [this](https://github.com/tramhao/termusic/issues/144) feature to be added I am also looking into adding this in the library menu, but it's not as critical if we the status bar shows it. I am also only starting this if/when I start working on that change.

Someone let me know if they have a good suggestion for how and where the icon should be placed in the playlist menu

PR #151, opened 2023-09-22
Feature show playing track

GaelicGabe:
This PR tries to do what is described in https://github.com/tramhao/termusic/issues/150

Summary:
* Places a configurable icon in front on the currently playing track title in the playlist
* Adds the song title to the progress bar
* Allows daemon to send updates to TUI
* Fixes bug that moves track selection back to old position right after playing a new track
* Fixes bug in debug mode where TUI will crash

It places a configurable indicator infront of the title inside the `playlist` component in all the layouts. The indicator is a string that defaults to `► `.

It additionally adds the title of the song to the progress component. Previously the title was only displayed for podcast episodes, now it is always displayed in addition to being first with its own `Title: ` header.

In order to solve a problem where the TUI would not track which the currently playing song track was when pressing next I've introduced the ability for the daemon to notify the TUI of updates. Currently the only update it can send is the `TrackChanged` update. This saves the TUI from polling the daemon and allows the TUI to be more snappy and responsive. This change allowed me to fix an existing bug where if you change track (for example by pressing `l` in the playlist view) and then quickly move your selected track the TUI will move your selection back.

Finally I fixed a bug that would cause the `tui` to crash if it was compiled in debug mode. The reason the crash happened was because, on init a `blocking::reqwest::Client` would be created which would in turn call some blocking request, however this was all done inside of an async context. Debug mode causes tokio to check for blocking requests in async contexts and panic.
The solution was to utilize `tokio::task::spawn_blocking`.


tramhao:
Due to a mistake of me, I have to revert all the PRs. Would you please submit this alone so I can merge it? It's very nice.

PR #154, opened 2023-10-09
Refactor daemon

GaelicGabe:
This PR refactors a lot of the code in the daemon.

* It gets rid of unnecessary `Arc<Mutex<Channel>>`
* Tries to keep clearer boundaries for separation of logic
* Moves the daemon tick driving loop to the main daemon event loop
* Removes the ctrl-C logic since it didn't do anything on either daemon or TUI
* Remove polling loops in a couple of places and replaced them with `async` alternatives
* Try to straighten out several confusing logic flows for daemon messages - for example playing a song required 3 `PlayerCmd` messages and 2 file synchronizations (1 write and 1 read). This was reduced down to 1 `PlayerCmd` message
* Try to simplify daemon logic and remove unnecessary state
* Allow for more general "next track" logic by having a `Iterator` be a generator for the next track - this lets us create whatever generator we want based on the loop mode
* Generally tries to improve small code bits and renames confusing names
* Leaves comments for potential future improvements


tramhao:
I believe this PR changed quite a lot of things. Would you please check if hotkey n for next track is working on your side?

hasezoey:
FYI: Most of this PR has now been re-implemented as of e608bc52d4928b007c55337bb19746ff88557126.
One notable exception is `NextTrackGenerator`, which has not been re-implemented yet.

re #170
re #173

GaelicGabe:
Waow, that's great! :) I'll check it out.

PR #168, opened 2023-11-24
fix(main): fix all build errors because of recent merges

hasezoey:
likely because of #155, #154 and #157

also removes the unused functions added by #154

#154 added some functions to outsource some functions (like checking for the daemon and starting it), but #157 changed it so that they depend on each-other (and it was merged so that the functions added by #154 were unused).

the full logic could likely be outsourced to another function (instead of individual functions)


tramhao:
Today I'm trying to catch up but right now the code in main is quite messy. Some function is not working. Like N for next song...

PR #171, opened 2023-11-24
Change actions used in workflows

hasezoey:
This PR changes the actions used in the workflows to non-deprecated versions

NOTE: please wait until the CI is finished with this one, so that we know there are no problems with the actions

another NOTE: the release run will have to be tested separately because it does not run for a CI (and i am unsure of the changes i have done for `actions-rs/cargo`)

PS: the cross/cargo steps could likely be combined by changing the matrix value to set which to use (instead of `cross: true` use `command: cross` and `cross: false` `command: cargo`), but for now i have gone with that

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,26 +28,16 @@ jobs:
       - uses: Swatinem/rust-cache@v1
 
       - name: Run cargo check
-        uses: actions-rs/cargo@v1
-        with:
-          command: check
+        run: cargo check
 
       - name: Run cargo test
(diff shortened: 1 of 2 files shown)


tramhao:
Currently the master branch have problems. Is it working on your side? I want to fix master branch first.

hasezoey:
well i could at least test the `build.yml` workflow (as you can see in the PR CI tests), nothing much changed, i only replaced deprecated actions with something that is not (behavior should not change - aside from the `cross` change)

PR #172, opened 2023-11-24
Change to use `flexi_logger` and add clap to server

hasezoey:
This PR basically had one goal: have logging to file, but this required some other things, in more detail:
- change from `lovely_env_logger` to `flexi_logger`
- disable logger dependency on crates that dont use it
- add clap config to `termusic-server` (basically a copy of `termus-tui`'s)
- add logging options to both cli structs
- add a `logging` module to both crates (server & tui)
- add dependency `colored`
- copy some extra functions (for config) from tui to server
- disable stderr logging in tui once starting UI
- disable screen clearing on TUI quit

The CLI arguments, config retrieving and logger module could likely be in a common crate to share code, but i have not done this there to "just have something working"

Note: the logger module is basically a copy from one of my projects
Note: i disabled the screen clearing because it is not necessary (and actually messes-up the screen) - because the alternate mode is used and so does not require any clearing (at least on UNIX-like terminal, i dont know about windows)

Some notes about the functions added:
- `--log-to-file` enables logging to a file, with the default being in `/tmp/termusic-CRATE.log`
- `--log-file` to set the location of the log-file (if specified automatically enables `--log-to-file`)
- `--log-filecolor` enables logging colors to the logfiles (is useful when doing `tail -f LOGFILE` or `less LOGFILE`)

PS: the default paths likely need to be refactored for actually getting the temporary directory on non-posix systems


tramhao:
After revert, this cannot be merged very easily. Meanwhile, currently I'm using RUST_LOG=trace termusic-server 2>log.log and it can log into file.
Do you think flexi_logger is better? Thanks.

hasezoey:
i will rebase and update it slightly (so that the defaults work on non-posix systems)

> `RUST_LOG=trace termusic-server 2>log.log`

that certainly is a option, but with this PR it provides a convenient argument to just start logging to a file, it also supports color logging (like the earlier examples for `tail -f` and `less`) and it can be passed over to the auto-started server too (and you are in full control on how the log lines are formatted)

hasezoey:
please try to check merge-ability before merging (this PR had NOT been rebased yet and i had not changed it for non-posix systems and the tests were failing (not because of clippy))

PR #194, opened 2024-01-21
Update dependency `souvlaki`

hasezoey:
~~This PR requires / is build on #193; do not merge this current PR first!~~

This PR updates dependency `souvlaki` (generic mpris), but i still wanna wait for some feedback on https://github.com/Sinono3/souvlaki/issues/46, because that version bumped the MSRV to 1.74

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3122,9 +3122,9 @@ dependencies = [
 
 [[package]]
 name = "souvlaki"
-version = "0.6.1"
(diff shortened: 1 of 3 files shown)


hasezoey:
`souvlaki` responded and fixed the error lowering the MSRV again (and also added a changelog and tested MSRV);
also updated to support event `SetVolume`, re #200

PR #196, opened 2024-01-22
Some Playback refactors

hasezoey:
This PR will look quite big, but is actually just some small changes, but with indentation / relocation changes, in more detail:
- fix mpv clippy & deprecation warnings
- fix mpv tests
- delete some unused files
- reduce indentation in some places
- combine some options / results to a chain (to reduce indentation of nested match blocks)
- rusty_backend(mod): relocate thread closure to own function
- rusty_backend(mod): de-duplicate creating a `MediaSourceStream` and appending to the sink
- rusty_backend(mod): exit loop (in thread) if channel errors with `Disconnected`
- rusty_backend(decoder): print non-fatal decoding errors to the log

if curious about the changes, i recommend going through each commit (i try to keep "unrelated" changes in their own commits)


hasezoey:
updated PR to drop 2 commits, which are better / differently handled in #205

PR #195, opened 2024-01-21
Update dependencies `gstreamer` and `glib`

hasezoey:
~~This PR requires / is build on #193; do not merge this current PR first!~~

This PR updates dependencies `gstreamer` and `glib`, this is a WIP because i had to disable some code, but i am not sure what it did, also because of a deprecation function that is still in use:

```txt
warning: use of deprecated associated function `glib::main_context_channel::<impl glib::MainContext>::channel`: Use an async channel, from async-channel for example, on the main context using spawn_future_local() instead
   --> playback/src/gstreamer_backend.rs:128:47
    |
128 |         let (main_tx, main_rx) = MainContext::channel(glib::Priority::default());
    |                                               ^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: `termusic-playback` (lib) generated 1 warning

```

PR #197, opened 2024-01-23
Refactor to support backend selection at runtime

hasezoey:
This PR refactors everything so that the backends can be compiled-in at the same time (making catching compile errors and testing a lot easier) and allow selection (cli argument) at runtime. in more details:
- change `Cargo.toml` features to use `dep:` prefix to remove implicit features
- change `server::Cargo.toml` to specify `playback` features for the same features (otherwise it need to be specified project-wide to work)
- add new feature `all-backends` to compile all backends at the same time, instead of having to least each one (small helper)
- add common function (like `gapless`, `set_gapless`, `skip_one`, `position.lock` and `enqueue_next`) to the `PlayerTrait` instead of invoking it on the backend directly (required for runtime selection)
- rename backends to be consistently named (`GStreamer -> GStreamerBackend`, `rusty_backend::Player -> rusty_backend::RustyBackend`)
- allow backends to be chosen at runtime
- implement cli arguments for backends, see Notes *1
- remove `PlayerCmd::DurationNext`, see Notes *2

Notes:
- *1: the server and tui both have a `--backend` added, where the tui passes it on to the server, but unlike the server the tui has no feature requirements (and so currently makes the error bad if specifying a backend that the server is not compiled for)
- *2: `PlayerCmd::DurationNext` had to be removed because only backend `rusty_backend` set it, and nothing (from what i can tell) actually ever read it / made use of it (i only found that the next complete `Track` would be given to a function to get the duration)

This PR will need changes if #196 should be merged first (otherwise #196 will need changes if this PR is merged first)


hasezoey:
rebased to fix conflicts with #195

hasezoey:
just noticed something that still needs to be updated (reminder for later): README table of features

hasezoey:
also updated the `Makefile` to reflect the changed features
in addition, updated the `post` script to install to `CARGO_HOME/bin` instead of the static `~/.local/share/cargo/bin`

tramhao:
DurationNext is used to enable gapless play. Did you test this PR with gapless playback? I use an album the wall to test it.

hasezoey:
> DurationNext is used to enable gapless play. Did you test this PR with gapless playback? I use an album the wall to test it.

i had always gapless play enabled when doing the PRs, and i didnt see any weird behavior; but i will test again after rebasing thanks to the merge of #196.
also note that i never actually saw any difference between gapless enabled play and not enabled play

hasezoey:
rebased to fix conflicts with #196 

> DurationNext is used to enable gapless play. Did you test this PR with gapless playback? I use an album the wall to test it.

i did some more reading on what gapless actually is and how it works, and i came to the conclusion that i cant *fully* test it, as i dont have any song (that was not previously in another format) in a format that is lossless or supports gapless from spec (like opus, unlike mp3).
but aside from that, i dont see why `DurationNext` would be necessary, as the only place it is actually used is to set the new `total_duration` on restart playing after `state == stopped && playlist == empty`, which is already handled by the song actually starting to play.

only read / used in:
https://github.com/tramhao/termusic/blob/20fc2e0ed9d6c4b1a7fab53452ca56c6a179ee80/playback/src/lib.rs#L190-L193

tramhao:
I don't remember the detail but the whole idea is, when gapless is enabled, the next song will be add to queue, and the queue will play next track when current one finish. I do have several files downloaded from website to test it. 
[gapless-sweep_part1_iis_1.zip](https://github.com/tramhao/termusic/files/14334459/gapless-sweep_part1_iis_1.zip)
[gapless-sweep_part2_iis_1.zip](https://github.com/tramhao/termusic/files/14334460/gapless-sweep_part2_iis_1.zip)
please change the extension to m4a. When gapless enabled, there is no noticeable pause.

hasezoey:
> I don't remember the detail but the whole idea is, when gapless is enabled, the next song will be add to queue, and the queue will play next track when current one finish.

you just gave me a idea; i think `DurationNext` was meant to store the duration of the next track while the current one is still playing without changing the `total_duration`(current playing songs duration).
but if this is the case, then it is already broken on master, and this PR did not change this behavior, and so will likely need further refactoring to fix. (will open a issue)

i also tested your files, i tested both in master and this PR, i dont see any differences; as in there is a noticeable "audio flicker" regardless of gapless enabled / disabled on my system on change from 1 to 2, but not from 2 to 1

hasezoey:
i did some testing and from what i said earlier:

> you just gave me a idea; i think DurationNext was meant to store the duration of the next track while the current one is still playing without changing the total_duration(current playing songs duration).
> but if this is the case, then it is already broken on master, and this PR did not change this behavior, and so will likely need further refactoring to fix. (will open a issue)

actually broke because of #196, i am not fully sure why yet, but will keep investigating

PR #198, opened 2024-01-24
Add a Currently Playing Symbol

### LINKED ISSUE #150: `playing` icon in front of currently playing track and currently playing track in status bar

hasezoey:
This PR adds a symbol to the currently playing track in the playlist:
currently_playing_symbol

This PR builds on the original commit of 5b2b5cf24d6c6ba9d4ce6e6d5e9ed7e61aef51d8 (part of #154), but enhances it so that it is actually configurable (like the other symbols)

Note: i am personally not happy with:
- the way it gets added and needs a full playlist re-sync to remove it from one title and add it onto another (instead of operating on the table directly)
- the way it looks / where the symbol is placed; it looks like it is part of the track name, but i didnt / dont know how to put it somewhere like the highlight symbol (cursor)

as for the location the config where this field is placed, it is right under `Playlist Highlight Symbol` (in the playlist column), because i didnt know a better place to put it

fixes #150

### DIFF
--- a/lib/src/config/theme.rs
+++ b/lib/src/config/theme.rs
@@ -129,6 +129,7 @@ pub struct StyleColorSymbol {
     pub lyric_background: ColorTermusic,
     pub lyric_border: ColorTermusic,
     pub alacritty_theme: Alacritty,
+    pub currently_playing_track_symbol: String,
 }
 
 impl Default for StyleColorSymbol {
@@ -151,6 +152,7 @@ impl Default for StyleColorSymbol {
             lyric_background: ColorTermusic::Reset,
             lyric_border: ColorTermusic::Blue,
             alacritty_theme: Alacritty::default(),
+            currently_playing_track_symbol: "►".to_string(),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


hasezoey:
@tramhao do you maybe know of a better way than the current playlist re-sync on title update? (for dynamic parts like the currently playing symbol) hopefully without having to re-implement the List widget

https://github.com/tramhao/termusic/pull/198/files#diff-aeea1c57af4800d34c38dfd04c90f07c502d1d2fb653c9398ecc5ea5bafc126dR844-R847

PR #199, opened 2024-01-24
Refactor some channel related types in playback

hasezoey:
This PR expands on a297816754ab4672752f3d67a484bdc61bf32674 (part of #154), which refactors channels for `PlayerCmd`, which includes:
- making a alias for the type, so it is easily used and referenced across many places
- remove `Arc<Mutex<` around the tokio channels, because they themself are `Send + Sync + Clone` (on both receiver and sender)

This PR will very likely conflict with ~~#196~~ and #197


hasezoey:
updated to resolve conflicts with #196

tramhao:
Some conflicts after merging previous prs.

hasezoey:
rebased to solve conflicts with #197 (and maybe others)

PR #205, opened 2024-01-27
Rusty decoder updates: add `mkv` + slight refactorings

hasezoey:
This PR basically adds symphonia feature `mkv` and updates all surrounding code to support it, and some refactors which annoyed me along the way (like changing `eprintln` to `error`), in some more detail:
- feat: add symphonia's `mkv` support
- fix: use first valid codec instead of always relying on `default_track` (which may be a unsupported stream)
- fix: skip all packets which are not for the current track (which made playing files with both video and audio, or multiple audio tracks impractical)
- refactor: replace `eprintln` and `println` with appropriate counterparts from `log`

This PR will conflict with #196 and maybe #197 and #199


hasezoey:
i forgot to note (more explicitly) that even though now symphonia / rust backend supports `mkv`, `lofty` does not, meaning all things that rely on it (like TUI playlist data) do not properly work (no duration indication, no title, no album, no artist and no cover)

hasezoey:
rebased to fix conflicts from #196

tramhao:
some conflicts happens.

hasezoey:
rebased to fix conflicts with #197 (and maybe others)

Issue #192, opened 2024-01-21
gstreamer (gst) TUI & Sever desync when song ends


hasezoey:
When having 2(or more) songs in the playlist and the first song is playing, but is fast-forwarded to the end, the server starts to play the next song, but the TUI's playlist (at least according to the Lyrics's title name) it is still the old song, and when that next song finished, it will loop (infinitely).
This only affects backend `gstreamer`, not `rusty_backend` or `mpv`.

Natural (non-seek) song ending is also affected, but the server will continue to next song and not loop (unlike the fast-forwarded problem where the song will loop)

Note that "Next song(N)" is not affected and actually fixes the de-sync (but still will re-play the already playing song)

Playlist mode: `playlist`
backend: `gstreamer` (feature `gst`)

<details>
<summary>some screenshots</summary>

![before_seek](https://github.com/tramhao/termusic/assets/10911626/ce240235-98d7-4a3c-aa24-e577dcfe91bc)
![after](https://github.com/tramhao/termusic/assets/10911626/44ecb1a7-7a95-4d76-84bb-97874cfd56f0)

</details>

PR #213, opened 2024-02-23
Gst updates

hasezoey:
This PR does a variety of thing to the `gst_backend`, most notably: fixing #192 and applying https://github.com/tramhao/termusic/commit/f4ed6659baaad449ad1b40c58760db65dbcb2d21#commitcomment-138909846, in more detail:

- apply https://github.com/tramhao/termusic/commit/f4ed6659baaad449ad1b40c58760db65dbcb2d21#commitcomment-138909846 so that gst events actually get send again
- replace glib warnings / errors with rust log ones
- remove only written-to (and otherwise unused) property `duration`
- remove only written-to (and otherwise unused) property `paused`
- clean-up some commented-out code
- allow volume above 100 (because gst allows it) (currently gets clamped somewhere else on set, but reads higher values from config just fine)
- change `is_paused` to check against specific states and log unusual states
- have EOS events in "gapless" mode (thereby fixing #192)

current downside to the implementation of the fix for #192 is that if you hold down `n`(skip one), then the key-repeats will get buffered (unlike in rusty; via unbounded channel) and gst-backend may decide to just stop playing anything (no crash, no further investigation for now); note that this buffering behavior also exists without this pr

### DIFF
--- a/playback/src/gstreamer_backend.rs
+++ b/playback/src/gstreamer_backend.rs
@@ -71,9 +71,9 @@ impl GStreamerBackend {
     #[allow(clippy::too_many_lines)]
     pub fn new(config: &Settings, cmd_tx: crate::PlayerCmdSender) -> Self {
         gst::init().expect("Couldn't initialize Gstreamer");
-        // let ctx = glib::MainContext::default();
-        // let guard = ctx.acquire();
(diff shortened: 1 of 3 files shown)


tramhao:
Thanks so much!

PR #216, opened 2024-03-01
Refactor to use and store `Duration` instead of `i64`

hasezoey:
This PR refactors "all" places (`PlayerTrait`, Backends, gRPC, TUI) to use `Duration` instead of `i64` (or even `u32` or `i32`).
There are some tangential things i have done:
- change mpv to listen on `Double`(`f64`)s instead of `Int64`(`i64`) (way more precision than just a second)
- add some doc-comments to `lrc`
- change music-player-service's shared state to be separate from the response (easier to use)

@tramhao please let me know if everything as `Duration` is fine, or should we use something like gst's `u64` which is nano seconds, or mpv's `double`(`f64`) storing(or at least for transfer) seconds?

### DIFF
--- a/lib/src/songtag/lrc.rs
+++ b/lib/src/songtag/lrc.rs
@@ -68,10 +68,13 @@ const EOL: &str = "\n";
 
 impl Lyric {
     // GetText will fetch lyric by time in seconds
-    pub fn get_text(&self, mut time: i64) -> Option<String> {
+    pub fn get_text(&self, time: Duration) -> Option<String> {
         if self.unsynced_captions.is_empty() {
             return None;
         };
+
+        #[allow(clippy::cast_possible_wrap)]
+        let mut time = time.as_secs() as i64;
 
         // here we want to show lyric 2 second earlier
         let mut adjusted_time = time * 1000 + 2000;
--- a/lib/src/songtag/lrc.rs
+++ b/lib/src/songtag/lrc.rs
@@ -126,11 +126,13 @@ impl Lyric {
         Some(index)
     }
 
-    pub fn adjust_offset(&mut self, time: i64, offset: i64) {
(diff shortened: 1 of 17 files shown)


tramhao:
I remember one place I use i64 by purpose. When seeking, if it's less than 0, it'll set the track to 0 secs. If u64 is used, it'll be an overflow.

hasezoey:
> I remember one place I use i64 by purpose. When seeking, if it's less than 0, it'll set the track to 0 secs. If u64 is used, it'll be an overflow.

for whats its worth, i didnt find a place where it would matter that the *progress report* is a positive-only value, the only place i noticed where it *could* be negative was in mpv (something along the lines of `-0.0001231` at track start), so i have some clamping to `0` there.

seeking *calls* (requests) still use `i64` as they need to indicate which direction to seek:

https://github.com/tramhao/termusic/blob/d3efb1355810fa132bc98c9e2ea2d396de1dc581/playback/src/lib.rs#L688

**except** `seek_to`, which is absolute in the track (but hadnt been changed in this PR) (which actually in all implementations also just seeks in seconds):

https://github.com/tramhao/termusic/blob/d3efb1355810fa132bc98c9e2ea2d396de1dc581/playback/src/lib.rs#L690

maybe in the future i will refactor `seek` to also use Duration to have the ability to seek relatively with more precision than seconds, but use something like gst is doing and wrap Duration in a `SignedUnsigned` enum:

[Signed in gstreamer](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/ffad1188b915b4181032eddb84a7e161a2acf7bc/gstreamer/src/format/signed.rs#L17-20), [implementation with ClockTime](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/ffad1188b915b4181032eddb84a7e161a2acf7bc/gstreamer/src/format/clock_time.rs)

PR #218, opened 2024-03-02
Refactor `PlayerTrait`'s volume

hasezoey:
This PR refactors `PlayerTrait`'s volume functions to take / return `u16`, including all backends and config, in more details:
- store volume as `u16` on all backends instead of `i32`
- store volume as `u16` in the config (which makes configs invalid which have negative volumes or way too high of volumes; which requires manual editing to even achieve)
- change proto definition to use `uint32` (no sizes below 32 bits are available in protobuf)

Why `u16` instead of something else like staying with `i32` or using `f32`?
I think `u16` is the best choice for the current spec, because:
- volume is never negative so signed values (`i*`) should fall out (as half would never be used)
- we are currently using volume as percent (both in tui, settings and config), so changing to floats (without the * 100 scaling) would be heavily breaking (it would likely also incur a incompatibility with existing configs)
- as for higher / lower bit types, `u8` *would* fit and allow to go way higher than we currently support (0 to 100), up to 255, but the backends also *could* go higher than that (even though it would likely not be practical), and anything higher than `u16` would likely be way overkill and so wasted space.

@tramhao is `u16` fine, or should it be changed to `u8`, or should a completely different type be used?

### DIFF
--- a/lib/src/config/mod.rs
+++ b/lib/src/config/mod.rs
@@ -290,7 +290,7 @@ pub struct Settings {
     pub max_depth_cli: usize,
     pub player_port: u16,
     pub player_loop_mode: Loop,
-    pub player_volume: i32,
+    pub player_volume: u16,
     pub player_speed: i32,
     pub player_gapless: bool,
     pub podcast_simultanious_download: usize,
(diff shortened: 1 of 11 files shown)


tramhao:
I think u16 is fine for volume. Thanks so much!

PR #224, opened 2024-03-04
TUI: wait until connected (and other conditions) instead of static amount of sleep

hasezoey:
This PR changes the TUI startup so that it waits until connected or error, or server died or timeout, instead of a static amount of sleep.


tramhao:
I merged it but have conflicts. Now my computer cannot compile so would you please push the PR again?

PR #229, opened 2024-03-09
Refactor `ueberzug` instance

### LINKED ISSUE #202: Windows: update photo error: Failed to run Ueberzug: program not found
Ueberzug doesn't even exist on Windows...

This bug's very annoying, happens on any view changes, and requires to hit `esc` each time

hasezoey:
This PR refactors how `ueberzug` is handled in error cases and windows, in more detail:
- dont re-try starting `ueberzug` on a permanent error (like binary not found)
- disable `ueberzug` even if feature `cover` is enabled on `windows`
- some general de-duplication in `ueberzug` instance

fixes #202 

PS: i do not know if the `not(target_os = "windows")` also affect mingw as i didnt find any documentation if mingw counts as that `target_os`

### DIFF
--- a/lib/src/ueberzug.rs
+++ b/lib/src/ueberzug.rs
@@ -5,6 +5,7 @@ use std::process::Child;
 use std::process::Stdio;
 use std::sync::RwLock;
 
+#[derive(Debug)]
 pub struct UeInstance {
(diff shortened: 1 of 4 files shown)


tramhao:
Under mingw64 ueberzug works with bugs. Maybe close the feature for now.

PR #231, opened 2024-03-11
Update uerberzug references in README

Porkepix:
Ueberzug original repository was abandonned by its author. It was forked under a new organization for some time which is the new point of reference for distros.
Ueberzugpp is a drop-in replacement written in C++.

As those were mentioned twice in README, I hesitated if first occurrence would need links too, and could also write it something like `ueberzug[pp]` instead of the way I did.

I was not sure if README was the place to mention that `ueberzugpp` was a drop-in replacement, so I opted for not write it there. I can update those depending on feedbacks, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ You can optionally install [yt-dlp](https://github.com/yt-dlp/yt-dlp/) and [FFmp
 
 #### Album cover support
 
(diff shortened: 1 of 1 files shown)


tramhao:
Thanks for the updates!

PR #247, opened 2024-03-20
Rusty Radio icy metadata cleanup and fixes

### LINKED ISSUE #242: rusty backend network download for radio open *2* duplicate streams
Rusty backend currently opens 2 streams, one for the actual audio data and one for audio data + icy metadata, meaning there are 2 active streams doing the same thing, but just doing one audio data + icy metadata would mean filtering out the metadata as symphonia does not directly handle it.

stream 1, used for symphonia:
https://github.com/tramhao/termusic/blob/04c6ff5040769247d24c9c06d2dde67355852472/stream/src/http.rs#L46-L50

streams 2, used for metadata:
https://github.com/tramhao/termusic/blob/04c6ff5040769247d24c9c06d2dde67355852472/stream/src/http.rs#L72-L76

maybe related #238

hasezoey:
This PR is a continuation of 85508bd1074ea38f8fb0e47c0a236bf5862f08e5, which had been merged prematurely, in more details:
- move icy-metadata handling to own file and module
- use header `icy-description` if available as initial title, until first title is parsed
- add fallback text for `StreamTitle='';` instead of empty `Currently Playing: `
- handle not having any icy-metadata
- remove `termusic-stream` as it is now unused in favor of `stream-download` directly
- add some icy-metadata resource comments (test streams and protocol definitions)

fixes #242


tramhao:
Thanks so much!

PR #249, opened 2024-03-20
QuitPopup & ConfigSavePopup: allow confirming choices with "Y" and "N"

### LINKED ISSUE #108: Allow to confirm we want to quit with Y and N
Just small usability improvement of something that we can commonly find elsewhere: when there's the popup asking if user is sure they want to quit, allow to use 'n' for No and 'y' for Yes (preferably both capitals or not).

hasezoey:
This PR adds keybindings for `Y` and `N` to the `QuitPopup` to confirm the choices.
Please note that i am not proficient in tuirealm, so i dont if i set it properly (it looks quite verbose to me)

fixes #108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Feat(server): for backend `mpv`, switch to use `libmpv-sirno` and use mpv API 2.0.
 - Feat(tui): add a "currently playing" symbol to active track in playlist.
 - Feat(tui): add search function for Podcast Episodes.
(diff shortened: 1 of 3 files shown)


hasezoey:
as a addition, i also added the same behavior to the config editor's "confirm" dialog

tramhao:
Yes, it's not very straight forward to add key bindings in tuirealm. Consider add to help as well. It's best to show everything without referring to other documentation.

PR #253, opened 2024-03-21
refactor(lib::track): change "media_type" to be non-optional

hasezoey:
This PR changes `Track`'s `media_type` to not be option, as it has not been used optionally. also a track should likely always identify what type it is.


hasezoey:
Note: i dont know why the mac test randomly failed, maybe try to re-run it

PR #282, opened 2024-04-04
fix(playback::rusty): clear already decoded sample buffer after seek

hasezoey:
This PR changes the rusty backend to clear its sample buffer after a seek.

I dont actually hear a audible difference, but i think it is actually more correct to not play the samples from before the seek.

@tramhao what do you think?


tramhao:
Sure, there is no need to play them.

PR #313, opened 2024-05-05
chore(typo): improve localization

MyyPo:
Improves some messages and error messages that I encountered in the app or greped.


tramhao:
There are some clippy warnings. Would you please modify and rerun the tests? Thanks.

PR #329, opened 2024-06-01
fix(lib::config::key): change "BindingForEvent" ser/de to be independent of lower lib representation

hasezoey:
This PR updates the current config key `BindingForEvent` config layout to be independent of lower library representation.
This was necessary because tuirealm 1.9 upgraded to bits 2.x, which changed its ser/de layout, which would break existing configs for termusic.

re #237

~~Will very likely conflict with #319~~

### DIFF
--- a/lib/src/config/v1/key.rs
+++ b/lib/src/config/v1/key.rs
@@ -176,9 +176,263 @@ impl Keys {
     }
 }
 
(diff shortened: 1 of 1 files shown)


tramhao:
There is no fix button on my side. Would you please fix the conflict? Thanks. @hasezoey

PR #338, opened 2024-06-16
Database Library updates

hasezoey:
This PR refactors the music library database handling by making it more consistent and easier to reason with, some more details:
- move database module to own directory and extract out the TrackForDB and migration code
- change insertion to be handled consistently and with less allocations
- change to not have any panics in the database code, upper code should handle that
- add doc comments
- remove intermediate vector in `need_delete`
- variable renames to be easier to understand
- `*_last_position`: refuse tracks without a `name` as that would not actually find / set the correct things

PS: this PR is marked as a DRAFT as the CI will fail unless #336 and #337 are merged first (and this PR rebased)


tramhao:
So it's ready to merge?

Issue #358, opened 2024-08-15
[Request] Bump release to `0.9.1` to fix packaging with rust `1.80.0`


VuiMuich:
It would be nice if you could release a minor bump to `0.9.1` in order to get the `time` crate dep update released for package maintainers (e.g. on nixpkgs).
Of course it is possible to cherry pick the relevant commit and patch packages so they build with rust `1.80.0`, but being able to use the latest release is always nicer :grin:

tramhao:
Sound good to me.

VuiMuich:
Nice, thanks.
So I will change the `nixpkgs` PR to a version update as soon as you have released the version update.

tramhao:
Released V0.9.1, but with some problems https://github.com/tramhao/termusic/issues/259.
Probably you'll need to configure some feature flags in nixpkgs, as the default feature flag for termusic-playback cannot compile. Thanks.

VuiMuich:
It seems to build fine in nixpkgs even without the feature flags, but I get this error in the checkPhase:
```
error: builder for '/nix/store/0afnkjrw4hzfqzi6d7skcds7cfxl16lb-termusic-0.9.1.drv' failed with exit code 101;
       last 10 log lines:
       > ++ cargo test -j 4 --profile release --target x86_64-unknown-linux-gnu --offline -- --test-threads=4
       > warning: `/build/.cargo/config` is deprecated in favor of `config.toml`
       > note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
       > warning: `/build/.cargo/config` is deprecated in favor of `config.toml`
       > note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
       > error: the listed checksum of `/build/termusic-0.9.1-vendor.tar.gz/soundtouch-ffi/src/lib.rs` has changed:
       > expected: 49b1a2fa7698edd0b89328daff83a96284c1874bb1b35a06e17aaf6194b035d0
       > actual:   b367ac09c56ffb5751ee62afbf46ddd23c65adcff8812e702ebc59bc7e44dd6a
       >
       > directory sources are not intended to be edited, if modifications are required then it is recommended that `[patch]` is used with a forked copy of the source
```
This might be a problem of nixpkgs very well, since when skipping the checkPhase the binary seems to run without any issues (just tried playpack of a track and `playerctl next` and `playerctl play-pause`).

tramhao:
[...]

Issue #259, opened 2024-03-23
Failed to release V0.8.0


tramhao:
Before I publish, I use
```
cargo hack check --feature-powerset --depth=1
```
and 
```
cargo hack check --feature-powerset --depth=2
```
to verify if each feature can compile along and can be combined. 
However, right now rusty,mpv,gstreamer and all-backends are exclusive. Many errors happened. I have to yank the version published to crates.io and revert the V0.8.0 tag.

How to combine these options?

tramhao:
Also the release ci give error:
```
<br class="Apple-interchange-newline">
Invalid workflow file: .github/workflows/release.yml#L1The workflow is not valid. .github/workflows/release.yml: Unexpected tag '!matrix.cross'
--


[Invalid workflow file: .github/workflows/release.yml#L1](https://github.com/tramhao/termusic/actions/runs/8402646915/workflow)
The workflow is not valid. .github/workflows/release.yml: Unexpected tag '!matrix.cross'
```

hasezoey:
> to verify if each feature can compile along and can be combined.
> However, right now rusty,mpv,gstreamer and all-backends are exclusive. Many errors happened. I have to yank the version published to crates.io and revert the V0.8.0 tag.

the likely problem is that crate `server` requires to set features in crate `playback`, but rust has no conditional to enable code based on a feature enabled in a lower crate and needs it enabled in the current crate too (ie you cannot have `mpv` enabled in `playback`, while feature `mpv` is not enabled on `server`).
in addition i made it a compile-error if no backends are compiled in (from my reading of `cargo-hack --feature-powerset` it tries to run with `--no-default-features` and so may end up running without any backends)

hasezoey:
if you really want to compile with no default features (ie no backends selected) and / or to have the `server` & `playback` consistencies a little better, i could refactor it (similar to how `cursive`'s backends work via boxed dynamic trait objects)

tramhao:
Maybe we could leave it for now. No need to change now. As long as conpiles are ok, I could live with that.

tramhao:
Here comes an awkward situation. When I publish termusic-playback, the default option is no backend, so I have to add "rusty" as default backend.
Then when I publish termusic-server,  the default option of playback will lead to uncovered pattern in line 288, and I have to change it back.
Magically, the installation by cargo install works. I really don't know why.
[...]

PR #359, opened 2024-08-21
Move "no backend selected" compile error to sever

hasezoey:
This PR moves the "no backend selected" compile error to the server package instead of the playback package (and does a runtime panic instead).
This is because currently the TUI makes use of the playback package without needing any backends, which makes the release process bad.

re https://github.com/tramhao/termusic/issues/259#issuecomment-2301443314
re https://github.com/tramhao/termusic/issues/358#issuecomment-2301646206

PS: also update the Cargo.lock for the new version as it had not been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## ChangeLog
 
+### next
+- Unreleased
+- Change: move the "no backend selected" compile error to the "-server" package instead of "-playback" (no need to specify a feature when compiling "termusic"(tui) now)
+
 ### [V0.9.1]
 - Released on: August 21, 2024.
 - Change: enable `log-to-file` by default.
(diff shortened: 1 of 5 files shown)


hasezoey:
Note: i have tested that compiling `cargo build --package=termusic-server && ./target/debug/termusic-server /location --backend=rusty` and `cargo build --package=termusic && ./target/debug/termusic /location --backend=rusty` and running both does work without errors. (the tui connected to the running server instead of a new one)

> and does a runtime panic instead

with this PR there is now a runtime panic instead when trying to use a backend in `Backend`, though there could be some other options (which i kinda like less unless specific use-cases should come up):
- add a Dummy backend (we currently dont need one for tests, would either be always compiled in and be dead-code or require relatively many flags)
- change all functions in `Backend` to return a result instead
- only compile-in `Backend` if any backend is selected

PR #362, opened 2024-08-22
docs(readme): update instructions for Arch Linux

orhun:
`termusic` is now in the official repositories: <https://archlinux.org/packages/extra/x86_64/termusic/> 🥳


tramhao:
That's great!

PR #381, opened 2024-10-29
refactor(lib): remove special MPEG tag handling

Serial-ATA:
MPEG files were being read twice to get the language and description for USLT frames. Since Lofty [0.20.0](https://github.com/Serial-ATA/lofty-rs/releases/tag/0.20.0), `TagItem` is able to retain both of those fields.


hasezoey (approved):
LGTM

PR #384, opened 2024-11-01
Refactor songtag LRC handling

hasezoey:
This PR refactors many parts of the `lib::songtag::lrc` module and touches upon some related lyric things, in more details:
- add tests for LRC parsing and writing
- fix writing milliseconds to LRC format (previously the completely wrong value was used)
- refactor a bunch of stuff to require less heap (string) allocations
- refactor a bunch of stuff to do less (repeated) searching
- general addition of doc-comments
- cleanup unused code / variables / properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Feat(server): on rusty backend, enable `aiff` codec support.
 - Fix: change default config ip address to `::1` instead of `::` (any old values on windows will need to be changed manually)
 - Fix: check for other tag types instead of just the primary tag type (for example a wav file with riff metadata instead of id3v2 would not get metadata)
-- Fix: correctly write LRC milliseconds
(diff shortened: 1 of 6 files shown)


hasezoey:
I did some more changes:
- add tests for `merge_adjacent`, `adjust_offset` and `get_text`
- change `Lyric::get_index` to not use a artificial 2 second offset anymore
- fix `Lyric::adjust_offset` inverting the given offset when below 10 seconds (ie `Lyric::adjust_offset(_, 1000)` would give `self.offset = -1000` instead of `self.offset = 1000`), this didnt seem intentional, but if it is we should add a comment
- change from casting some values to using `try_from`
- use milliseconds directly instead of `as_secs`
- some style changes
- less cloning again
- change tui `no lyrics available` message to be the else case instead of basing it on another value
- add a note why `Lyric::get_text` uses a artificial 2 second offset

PR #412, opened 2025-01-14
Feature/improve searching options

Refael10ru:
add an option to specify what you search by
when "quick searching" in the database/playlist

The search options (for now) will be by album/artist/title.

I will need some guidance as to what are the standards of Rust code since this is my first time contributing using thing this language. 

+ I opened it early since I'm not sure how it is best to change the API between the client and the server.

### DIFF
--- a/lib/src/library_db/track_db.rs
+++ b/lib/src/library_db/track_db.rs
@@ -147,6 +147,10 @@ impl TrackDBInsertable<'_> {
     }
 }
 
+/// Defined for types which could be indexed.
+/// Was made to allow generalization of indexing/search functions.
+///
+/// the required functions are generally the metadata you would find in an mp3 file.
 pub trait Indexable {
     fn meta_file(&self) -> Option<&str>;
     fn meta_title(&self) -> Option<&str>;
--- a/lib/src/library_db/track_db.rs
+++ b/lib/src/library_db/track_db.rs
@@ -146,3 +146,70 @@ impl TrackDBInsertable<'_> {
         )
     }
 }
+
+pub trait Indexable {
+    fn meta_file(&self) -> Option<&str>;
+    fn meta_title(&self) -> Option<&str>;
+    fn meta_album(&self) -> Option<&str>;
+    fn meta_artist(&self) -> Option<&str>;
+    fn meta_genre(&self) -> Option<&str>;
+    fn duration(&self) -> Duration;
+}
+
+impl Indexable for Track {
+    fn meta_file(&self) -> Option<&str> {
+        self.file()
+    }
+    fn meta_title(&self) -> Option<&str> {
+        self.title()
+    }
+    fn meta_album(&self) -> Option<&str> {
+        self.album()
+    }
+    fn meta_artist(&self) -> Option<&str> {
+        self.artist()
+    }
+    fn meta_genre(&self) -> Option<&str> {
+        self.genre()
+    }
+    fn duration(&self) -> Duration {
+        self.duration()
+    }
+}
+
+impl Indexable for TrackDB {
+    fn meta_file(&self) -> Option<&str> {
+        if self.file == UNKNOWN_FILE {
+            return None;
+        }
+        Some(&self.file)
+    }
+    fn meta_title(&self) -> Option<&str> {
+        if self.title == UNKNOWN_TITLE {
+            return None;
+        }
+        Some(&self.title)
+    }
+    fn meta_album(&self) -> Option<&str> {
+        if self.album == UNKNOWN_ALBUM {
+            return None;
+        }
+        Some(&self.album)
+    }
+    fn meta_artist(&self) -> Option<&str> {
+        if self.artist == UNKNOWN_ARTIST {
+            return None;
+        }
+        Some(&self.artist)
+    }
+    fn meta_genre(&self) -> Option<&str> {
+        if self.genre == UNKNOWN_GENRE {
+            return None;
+        }
+        Some(&self.genre)
+    }
+
+    fn duration(&self) -> Duration {
+        self.duration
+    }
+}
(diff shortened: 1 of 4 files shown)


hasezoey (commented):
This looks good, though i am not on my test machine currently.

I would recommend to re-base this PR on my #407 changes and then run `cargo clippy` (and also `cargo fmt`) and see if there is anything to change (all warnings will need to be fixed to pass CI).

PS: if i understand this correctly, the database-component and the playlist-component basically had the search logic duplicated, which this PR de-duplicates, right?

tui/src/ui/components/database.rs
```diff
@@ -618,46 +619,64 @@ impl Model {
...
+                wildmatch::WildMatch::new(&search).matches(&title.to_lowercase())
+            } else {
+                false
+            };
+            let album_match: bool = if let Some(album) = record.meta_album() {
+                wildmatch::WildMatch::new(&search).matches(&album.to_lowercase())
+            } else {
+                false
+            };
+            if artist_match | title_match | album_match {
```
hasezoey, line 642:
```suggestion
            if artist_match || title_match || album_match {
```

Make it a logical OR instead of a binary OR

hasezoey (commented):
Works great, though clippy still needs to fixed, also you dont need to annotate types where not necessary (ie you can remove `: TableBuilder` and `: bool`).

Also now that it is searched by album, it should also list it as a column in the output, otherwise it might be confusing why a track is in there.

Refael10ru:
Another point that you may want to know is, 
that given whose changes searching for "Unknown artist" won't give you results anymore in my version of the search

lib/src/library_db/track_db.rs
```diff
@@ -146,3 +146,70 @@ impl TrackDBInsertable<'_> {
         )
     }
 }
+
+pub trait Indexable {
```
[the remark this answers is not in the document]
Refael10ru: done.
you can look at the relevant commit to see if it is good enough.

tui/src/ui/components/database.rs
Refael10ru: done

hasezoey (commented):
> Another point that you may want to know is,
> that given whose changes searching for "Unknown artist" won't give you results anymore in my version of the search

I think that is currently something we can ignore. This is because once doing #291 those fields will not be defaulted to some `Unknown` thing and instead be NULL / a Option, not requiring to be matched against the string constants to know if they are available or not.

lib/src/library_db/track_db.rs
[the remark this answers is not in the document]
Refael10ru: I will move it to the right spot
for some reason I though I saw other documentation placing it in the same spot.

Refael10ru:
Clippy doesn't report anything problematic

 
should I rebase over master again

Refael10ru:
Clippy doesn't seem to flag anything 
image

should I rebase it over master again?

Refael10ru:
Think we might be ready for merge after I rebase/do not rebase my branch back (depends on what you want)

I will implement defaults for searching by Album/Artist/Title in the next pr to keep them small.
As of the next pr, I am thinking between:
1) toggling options between the search types (Album/Artist/Title/ALL) or combinations
2) some defaults for what to search by depending on the originating state

hasezoey:
> Clippy doesn't report anything problematic

Good, though it will need to pass the CI before it is being able to merge. And as you are a first time contributor here, you will need your CI approved to be run, which can only be done by members with Write permissions (or Triage, i actually dont know), and i do not have either, you will have to wait for @tramhao.

> should I rebase over master again

I would recommend you rebase again, as currently you seem to have re-commited existing commits in the master branch (like `deps(tokio): update to 1.43`) and also re-commited things from #407 (like `style: fix 1.84 clippy warnings`) (and maybe even some other things?)

A rebase in this case should be as easy as (while being in the branch you want to rebase) `git rebase OtherBranch`.
Alternatively, you can rebase onto master again (make sure your master is up-to-date with the upstream master) and just merge #407 instead of rebasing onto that.

Refael10ru:
another thing that I may add in this PR is maybe try combining this search update
image
I didnt do that at the start because they have conflicting table layouts
and it would be a breaking change to Change it in a raw way

to solve it I can make a more general table builder something like
```rust
def build_table(data: Vec<Indexable>, show: citmap<4>) -> Table:
...
```

Refael10ru:
rebased over the post-merged master

hasezoey:
> another thing that I may add in this PR is maybe try combining this search update

Personally, i would recommend to put this in another PR.

lib/src/library_db/track_db.rs
```diff
@@ -146,3 +146,74 @@ impl TrackDBInsertable<'_> {
         )
     }
 }
+
+/// Defined for types which could be indexed.
+/// Was made to allow generalization of indexing/search functions.
+///
+/// the required functions are generaly the metadata you would find in an mp3 file.
```
hasezoey, line 153:
```suggestion
/// The required functions are generally the metadata you would find in an mp3 file.
```

tui/src/ui/components/database.rs
```diff
@@ -618,46 +619,62 @@ impl Model {
...
+                    table.add_row();
                 }
+                idx += 1;
+
+                let duration =
+                    termusiclib::track::Track::duration_formatted_short(&record.duration());
+                let duration_string = format!("[{duration:^6.6}]");
+
+                table
+                    .add_col(TextSpan::new(duration_string.as_str()))
```
hasezoey, line 653:
```suggestion
                    .add_col(TextSpan::new(duration_string))
```

I think this can be given directly now and avoid a clone.

Refael10ru: done

lib/src/library_db/track_db.rs
Refael10ru: done

tramhao:
Thanks a lot! @Refael10ru

PR #415, opened 2025-01-18
Feature/general table creator for quick searches

Refael10ru:
This pr is the second step in merging all the search functions that are used in the "quick search" menu into one.

this pr came to solve the different table layouts that search in the library have compared to database/playlist searches
and then make the library search use the general search logic. 

I again, opened the CR early since the type we use to represent the Table layout is a design choice I think I can't make by myself rn.

### DIFF
--- a/tui/src/ui/components/database.rs
+++ b/tui/src/ui/components/database.rs
@@ -629,6 +629,8 @@ impl Model {
     }
 
     pub fn update_search<'a, T: Indexable>(indexable_songs: &'a Vec<T>, input: &str) -> Vec<&'a T> {
+        // cr note: we could mabye make indexable_songs mut to avoid allocating another Vec
+        // or we could just allocate the full size of indexable to the new vec to avoid reallocations
         let mut filtered_records = vec![];
         let search = format!("*{}*", input.to_lowercase());
         for record in indexable_songs {
--- a/tui/src/ui/components/database.rs
+++ b/tui/src/ui/components/database.rs
@@ -628,10 +628,9 @@ impl Model {
         self.database_sync_results();
     }
 
-    pub fn update_search<T: Indexable>(indexable_songs: &Vec<T>, input: &str) -> Table {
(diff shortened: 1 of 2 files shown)


Refael10ru:
as per usual I make sure it passes `cargo clippy` checks

tui/src/ui/components/database.rs
```diff
@@ -628,10 +628,11 @@ impl Model {
         self.database_sync_results();
     }
 
-    pub fn update_search<T: Indexable>(indexable_songs: &Vec<T>, input: &str) -> Table {
-        let mut table: TableBuilder = TableBuilder::default();
+    pub fn update_search<'a, T: Indexable>(indexable_songs: &'a Vec<T>, input: &str) -> Vec<&'a T> {
+        // cr note: we could mabye make indexable_songs mut to avoid allocating another Vec
+        // or we could just allocate the full size of indexable to the new vec to avoid reallocations
+        let mut filtered_records = vec![];
```
hasezoey, line 634: At least if we keep the code using `Model::update_search` currently, then no, we cannot take `&mut Vec` and i dont think `mut &Vec` will be of much help avoiding reallocations. (see `Model::playlist_update_search`, which uses `self.playlist.tracks()` which only returns `&Vec<Track>`)

What we *could* do is make this function (and `build_table`) accept & return a iterator, this way we could avoid the allocations of another vec and likely avoid some multi-looping, though this would require a bit more of a refactor.

hasezoey (approved):
This PR Looks good to me. While you are at it, i would recommend to maybe put some documentation on the 2 functions modified in `database.rs`.
(Also remove the `cr note` before merge, right?)

Refael10ru:
Umm..
Didn't expect to see it merged when I came back to answer/fix the cr

tui/src/ui/components/database.rs
Refael10ru: Will move it to an iterator next pr
+
Still looking for a struct type to describe the colums of a table

PR #422, opened 2025-01-28
Feature/customiseble table creator

Refael10ru:
refactor to use iterators as asked in #415

make the table creator configurable
^ may need some advice on what is acceptable "API" for that


Refael10ru:
+ I would be glad if you know any good way to get rid of the code duplication of the Indexable implementation

hasezoey (approved):
Looks good to me, some doc-comments for the functions would be nice though.

Issue #449, opened 2025-03-11
UI crash when reopening Termusic quickly after closing it in WezTerm


Porkepix:
## Environment & Versions

- System: ArchLinux
- Termusic version: 0.10/master
- Rust version: 1.85.0

## Description of the Problem

I'm not sure about logs yet, the problem might not even come from Termusic but be a Wezterm's one, in which case I'll report to their side as I didn't see (for now) the issue on other terminals.

Issue arise when I opened correctly Termusic a first time, quit it properly (`Q > Yes`) and then reopen it quickly after closing it.

Most of the UI besides cover render at first, but then everything is frozen and hitting `ctrl-c` works to get a prompt appearing over the UI. It appears, from what I can see and guess that it prints on the terminal most of the UI, and then UI program crash while server continue to play its music in the background (and be reconnected to).

Additional notes:
Log only contains this (probably unrelated) message:
`[2025-03-11T19:00:26.675+01:00 WARN lofty::id3::v2::read]: Replaced frame with ID "TXXX" by a frame with the same ID`.

Curiously I'm not able to reproduce the issue on the other computer. Same distro, just one is running master with AUR while the other have stable package. Different music, too.

tramhao:
I think that is probably because the internal logic of termusic. This is a client-server program. While you close the app, maybe the server is still running for a while, and if you reopen the app, it'll not reopen the server because it's running. However, after several seconds, the server closed and client cannot communicate with server.

tramhao:
That is because by default termusic will kill termusic-server when quitting. Please change the configuration `Kill daemon when quit app` to `no`, and I think the problem should disappear.

hasezoey:
> So, I found both kill_daemon_when_quit = true in .config/termusic/config.toml and quit_server_on_exit = true in .config/termusic/tui.toml that might fit, I'm not sure what's different between them

FYI, as tramhao already explained, [since 0.9.1, the config got a V2 update](https://github.com/tramhao/termusic/blob/master/CHANGELOG.md#v091), splitting server and tui parts, you can safely delete the old `config.toml`, unless you run versions before 0.9.1.

> Seems that wezterm has some limitation to use HTTP ports?

To my knowledge, which kind of terminal is used does not affect http ports as that is not interacting with the terminal in any way.

> I'd have thought it'd have used UNIX pipes/sockets.

Not currently, see #433.

> Log only contains this (probably unrelated) message:
[...]

PR #461, opened 2025-03-27
fix(tui::ui::components::xywh::show_image): force viuer to only use the protocol we probed for

hasezoey:
This is necessary as viuer will always probe for all enabled protocols on each print until it finds a supported one, but because of likely improper implementation, it could cause a infinite loop of waiting for a response which never comes, freezing the TUI.

re #449

@Porkepix could you test this PR without `--disable-cover` or `--hide-cover` and with features `cover`(all) enabled across multiple TUI starts & tracks?

PS: this PR is marked as DRAFT until confirmed that it fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Feat(tui): add new cli option `--disable-cover` which actually disables all cover probing and behaves as if no cover features are enabled.
 - Fix(tui): set `ueberzug` command to `--silent`.
 - Fix(tui): force viuer to use the terminal image protocol that was probed at the beginning.
+- Fix(tui): add CTRL+C handler for when TUI key-reading is not active yet or not active anymore.
 
 ### [V0.10.0]
 - Released on: March 8, 2025.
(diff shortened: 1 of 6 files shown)


hasezoey:
Also tacked-on a ctrlc handler for the TUI, which is useful for when the TUI key-reading is not yet active or not active anymore. (like before starting the actual key handler / after shutting it down) or when something else takes key reading control, like viuer.

Funnily enough, a single CTRL+C press also exits the wezterm-sixel loop in viuer now, hence my implementation of needing a double-press within 2 seconds to actually exit from a CTRL+C.

hasezoey:
> Not sure how to test this one, and if you need this to be tested.

No need to test *that*, i just tested the ctrl+c handler against it to see if it would more gracefully exit if a similar issue should ever happen with viuer. (ie i ran it as if 1d00ffd59f8561fdd7f24c620bd910b33d178d3f was reverted)

> Also, cover isn't redrawn anymore after exiting ctrl + h which I believe used to be (but I'm not sure), so that cover is altered until another even ask for a new cover or this one to be redrawn.

As i already wrote in the original issue, this PR did not change anything regarding this.

---

Marking this PR as Merge-able as it is confirmed to fix (or at least workaround) the issue.

PR #467, opened 2025-04-02
Make rusty backend file decode / playback async

### LINKED ISSUE #191: Backend `rusty_backend` seems to not load enough data
In 4/5 runs of a title, the `rust_backend` has a `ALSA: underrun detected` (which is audible as stops), this may be because of my "slow" disk i have my music on, but using backend `mpv` just works without hiccups.
I have already tried investigating if i can somehow increase Buffer size of the title that is loaded, but i could not find a solution that actually works (or how symphonia / rusty_backend work); symphonia seems to have a hard lower limit of 1kB load and a max of 32kB, which on a HDD is not really enough

termusic version 43d75115d676fc6c74f5eb1bf2ee6ef23923e63d (#190 ) / 32f71c47ad1b2656fd0ab49704f8bec683761644 (current master)

hasezoey:
This PR changes the rusty decoder / playback to be async from each other, meaning that the decoder can keep the playback buffer full without first having to become empty and only *then* being able to load more data, which lead to #191.

Note that the ringbuffer is only enabled for local files which are seek-able (ie Music & Podcasts). Streams like radio already are async enough (they are async in terms of getting the raw data, not the decoded data).

Questions:
- [x] I have implemented my own buffer here, but after some more looking, it could be exchanged for [`bytes::BytesMut`](https://docs.rs/bytes/1.10.1/bytes/struct.BytesMut.html), though with that implementation, we would likely need to use more `unsafe` (in a safe way)
- [x] I have also implemented my own encoding / decoding of messages, but it would partially be exchanged for [`bincode`](https://docs.rs/bincode/2.0.1/bincode/)
- [x] I have some usages of `unsafe` in this PR, which i have checked to be safe, can they stay this way?

What do you think @tramhao?

Some future TODOs:
- Maybe tune some buffer sizes
- Change `AboutToFinish` time to run earlier (earlier loading of next track)
- Allow tuning buffer (and other) values in backends via the config

Fixes #191

PS: This PR is a DRAFT until some more testing can be done, though i have confirmed it working on my machine.

EDIT, as i forgot to include it originally, this work is based on the recommendation of symphonia: https://github.com/pdeljanov/Symphonia/discussions/260#discussioncomment-8276456


tramhao:
I don't like too many libraries so I tend to implement simple functions on my own and keep the imports to a minimum level. So it should  be no problem for your buffer. Unsafe code is ok as long as there is no memory leaks. Thanks.

hasezoey:
> Unsafe code is ok as long as there is no memory leaks.

Good, just as a FYI, the only newly added unsafe currently is re-interpreting a slice from the ValueType(i16) as bytes(u8) without copy or allocation.

> So it should be no problem for your buffer.
> I don't like too many libraries so I tend to implement simple functions on my own and keep the imports to a minimum level.

At least `bytes` is already a dependency of the TUI and the lib (and as a sub-dependency), maybe it would be OK to also add it for playback?

hasezoey:
Added a little more cleanup and documentation.
Additionally, i also added a benchmark for it to see how it performs.

With that benchmark, i have tested to see if `bytes` would behave any differently than the current implementation, and found to perform the same, but `bytes` had some annoyances when switching to it without refactoring too much. (Also the current implementation guarantees to never allocate anymore, whereas `BytesMut` *may* implicitly in some functions)

@tramhao I would say i am finish with this PR now (aside from maybe tweaking some values here and there), how does it run for you?

PS: linting fails now as rust 1.86 released with new lints again, will put in another PR to fix those.

hasezoey:
I think this PR is ready. Like mentioned earlier, it would be great to get some additional testing and maybe tweak some values here and there, but overall it works. (also just fixed some tests which depended on scheduling race conditions)

hasezoey:
FYI: i am aware that somehow async_ring may panic on some condition, in my case when trying to play mp3 files (i basically only tested flac as they were bigger) due to some kind of tokio runtime blockade.
Note that this only affects debug builds, will try to fix it soon.

```txt
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

PR #470, opened 2025-04-05
Some Dependency changes

hasezoey:
This PR removes unused dependencies, updates `stream-download` and switched out `futures` for `futures-util` to shed some other dependencies we dont use. (at least in the future)


hasezoey:
Update, since opening this issue, i again updated `stream-download` to shed some more unused dependencies.
Also disabled some features on `rss` we dont use and finally, updated all things to use `idna@1` in the `Cargo.lock`.

PR #473, opened 2025-04-12
chore(workflows/build): set extra cache key with rust version

hasezoey:
As with different rust versions, rust will always re-compile fully. And if there is only one key, the other run may fully need to re-compile.

Note that at current, the extra keys are only necessary for the `linux` tests, but in the future, if something changes on the other tests, it is available too.

PS: this is marked DRAFT as i wanna confirm the CI runs correctly before potentially merging.


hasezoey:
Forgot again that `Swatinem/rust-cache` already takes rust toolchain into consideration and that most of the time in the linux paths is spent on apt.

Issue #472, opened 2025-04-11
[Feature Request] Refuse to kill daemon if two or more termusic instances are active


shuntia:
## Describe what you want

Check before killing `termusic-server` if another instance of `termusic` is running

## Do you have already an idea for the implementation?

I have no idea of how `termusic-server` handles termination requests, but the server could check number of active clients, and reject the termination request if there is another client running.

hasezoey:
I also think this would be great, but this would mean actually implementing quitting in the server.
Currently the system works that if value `behavior.quit_server_on_exit` is `true`, then on TUI shutdown, it sends `SIGTERM` on unix systems and otherwise `SIGKILL` to the server process.

shuntia:
Wouldn't simple logic work?(I know there's a better way of detection, but using pgrep)
I don't know for sure, since I am a rather beginner.
`behavior.quit_server_on_exit` could be changed to an enum like `Always, Never, Default` to accommodate this. I honestly don't know how easy it is to do this, but it could be possible.

```rust
use tokio::process::Command;
#[cfg(unix)]
{
let pgrep=Command::new("pgrep").arg("termusic").stdout(Stdio::piped()).spawn().unwrap();
pgrep.wait().await;
let buf=String::new();
pgrep.stdout.unwrap().read_to_string(&mut buf).await;
if buf.chars().filter(|el|el=='\n').count()==1{
    killdaemon();
}else{
    exit(0);
}
}
```

hasezoey:
> Wouldn't simple logic work?(I know there's a better way of detection, but using pgrep)

Yes that (or translated) logic *would* work, though it would still be kinda fragile as it wouldnt quite know if those TUI instances are for a specific server (there could theoretically run multiple servers and multiple instances) and also would not work at all if the TUI and server run on different systems (or as different users).
Also if not running on a unix system, `SIGKILL` would still be used, which is not a graceful exit.

I had intended to rewrite that logic to just send a `Quit` Event and let the server decide if it wants to exit or not.

shuntia:
sometimes I forget that I have multiple clients open(That's on me), but I wanted to make sure that the other TUI won't crash if I blindly close the other one, since it's not really a graceful shutdown.
If I were to actually implement it, it would require the server to remember the number of the clients to be remembered by the server, or am I wrong?
If that's the case, I'll just scrap that PR and recreate a new one.

PR #477, opened 2025-04-18
refuse kill if another instance is running

shuntia:
This solves the issue:
#472 

This will:
1. on quit, probe for `termusic` instances at the same time `termusic` iterates over processes
2. count them if they are not spawned by `termusic` by looking up PID.
3. if there are two or more externally spawned processes, it will skip killing the server.

My code is not very elegant, so I'd like some simple feedback if possible. Thank you!

### DIFF
--- a/tui/src/ui/mod.rs
+++ b/tui/src/ui/mod.rs
@@ -159,17 +159,29 @@ impl UI {
         {
             let mut system = System::new();
             system.refresh_all();
+            let mut target = None;
+            let mut clients = 0;
             for proc in system.processes().values() {
-                let Some(exe) = proc.exe().map(|v| v.display().to_string()) else {
+                let exe = proc.name().to_str().unwrap();
+                if exe == "termusic-server" {
+                    target = Some(proc);
                     continue;
-                };
-                if exe.contains("termusic-server") {
-                    #[cfg(not(target_os = "windows"))]
-                    proc.kill_with(sysinfo::Signal::Term);
-                    #[cfg(target_os = "windows")]
-                    proc.kill();
-                    break;
                 }
+                let parentistermusic = match proc.parent() {
+                    Some(s) => system.processes().get(&s).unwrap().name() == "termusic",
+                    None => false,
+                };
+                if exe == "termusic" && !parentistermusic {
+                    clients += 1;
+                }
+            }
+            if clients == 1 && target.is_some() {
+                #[cfg(not(target_os = "windows"))]
+                if let Some(s) = target {
+                    s.kill_with(sysinfo::Signal::Term);
+                }
+                #[cfg(target_os = "windows")]
+                target.unwrap().kill();
             }
         }
 
--- a/tui/src/ui/mod.rs
+++ b/tui/src/ui/mod.rs
@@ -33,6 +33,7 @@ use futures_util::FutureExt;
 use model::{Model, TermusicLayout};
 use music_player_client::Playback;
 use std::time::Duration;
+use sysinfo::Pid;
 use sysinfo::System;
 use termusiclib::player::music_player_client::MusicPlayerClient;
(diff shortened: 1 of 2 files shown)


tui/src/ui/mod.rs
```diff
@@ -159,17 +159,29 @@ impl UI {
         {
             let mut system = System::new();
             system.refresh_all();
+            let mut target = None;
+            let mut clients = 0;
             for proc in system.processes().values() {
-                let Some(exe) = proc.exe().map(|v| v.display().to_string()) else {
+                let exe = proc.name().to_str().unwrap();
```
hasezoey, line 165: Preferably, this should not panic and instead just skip this process, please use something like `let Some(exe) = ` again like before the patch.

tui/src/ui/mod.rs
```diff
@@ -159,17 +159,29 @@ impl UI {
...
+                if exe == "termusic-server" {
+                    target = Some(proc);
                     continue;
+                }
+                let parentistermusic = match proc.parent() {
```
hasezoey, line 170: Consider renaming this variable to something like `parent_is_termusic`.

tui/src/ui/mod.rs
```diff
@@ -159,17 +159,29 @@ impl UI {
...
-                    #[cfg(not(target_os = "windows"))]
-                    proc.kill_with(sysinfo::Signal::Term);
-                    #[cfg(target_os = "windows")]
-                    proc.kill();
-                    break;
+                if exe == "termusic" && !parentistermusic {
+                    clients += 1;
+                }
+            }
+            if clients == 1 && target.is_some() {
```
hasezoey, line 178:
```suggestion
            if clients <= 1 && target.is_some() {
```

tui/src/ui/mod.rs
```diff
@@ -159,17 +159,29 @@ impl UI {
...
+            let mut clients = 0;
             for proc in system.processes().values() {
-                let Some(exe) = proc.exe().map(|v| v.display().to_string()) else {
+                let exe = proc.name().to_str().unwrap();
+                if exe == "termusic-server" {
+                    target = Some(proc);
                     continue;
+                }
+                let parentistermusic = match proc.parent() {
+                    Some(s) => system.processes().get(&s).unwrap().name() == "termusic",
```
hasezoey, line 171: This should likely also not panic.

hasezoey (changes requested):
Looks mostly good to me as a band-aid fix, though some style recommendations.

Like i said in #472, this is mostly a band-aid fix as there could be multiple clients that are not `termusic` or server and client may not even be on the same system (also there is the whole Signal thing).

Once again, i would be curious what your use-case of multiple clients currently is.

shuntia:
I sometimes forget that I have an instance of `termusic` running and blindly close that TUI, thus resetting my current track on the playlist. currently there's no option to keep the server alive without purging the extra instance with: `SIGKILL` or `SIGTERM` or re-configuring `termusic`.
I would just simply scrap this PR and create a Quit command if that's necessary.

shuntia:
I'll try rewriting the main code to create a new message type. in the meanwhile(while it is unlikely) I took into account your comment on the previous issue. It would save the PID of the server in `SERVER_PID` and priortize that over killing any other instance of `termusic-server`.

shuntia:
I'd need to read in way more if I want to do that(like probably create a service to track clients) but I don't think I understand enough.

hasezoey:
> I'd need to read in way more if I want to do that(like probably create a service to track clients) but I don't think I understand enough.

No problem, just wanted to write that would be the solution i would go for. (i also planned to that at some point, though had not worked on it yet)
In any case, this PR is handling it better than before (not that i have actually ran it yet, just from the general logic).

PR #478, opened 2025-04-19
feat(playback::rusty): change all sample types from i16 to f32

hasezoey:
This PR changes all places we interact with samples to use Sample-Type `f32` instead of the previous `i16`, reasons are (by order):
- in the next version of rodio, Sources will only work on `f32` (https://github.com/RustAudio/rodio/pull/700, possibly rodio `0.21.0`)
- symphonia decodes into `f32` in many formats to my knowledge (but later got converted by us)
- soundtouch already forces use of `f32` samples in the chain, if enabled
- audio systems (like pipewire, pulseaudio) basically also operate on f32 anyway

This means we just have to do less conversions and are better ready for the next rodio version.


hasezoey:
While i was modifying the Sink, i did some extra stuff:
- removed unused values
- removed duplicated modifiers (skippable & stoppable are the same in rodio 0.20.x)
- abstract away which speed modifier is used to not be in `Sink` anymore
- some small style changes

PR #479, opened 2025-04-22
Add support for Unix Domain Sockets

hasezoey:
This PR mainly adds Unix Domain Socket (or also referred as UDS) support. It is also chosen as the default for `unix` targets. Some notes:
- UDS is currently only available for `unix` targets in tokio & tonic, but windows (since some 10 version) also supports UDS, so it might come to there too
- UDS will be able to be parsed from the config regardless of target
- when migrating from v1 config, `HTTP` is chosen as the default
- if UDS is not available, `HTTP` is chosen as the default
- if UDS is not available, but set in the config, a runtime error will occur

Additionally this PR does, as it somewhat impacted the implementation:
- better gracefully exit instead of calling `::exit(0)` in a `Quit` message (allows tasks to exit gracefully)
- pause & stop the player before exiting. As without the `::exit()` call, the process would be kept around until the last audio source had been consumed

PS: this is a DRAFT as i wanted to know if it builds on other targets than `unix` or if i forgot some place

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### next
 - Change: updated MSRV to 1.82.
+- Change: support Unix Domain Sockets in addition to HTTP transport. (Note that UDS is only available for `unix` targets)
+- Change: change default transport protocol to be UDS on `unix` systems.
 - Change(tui): rename cli option `--disable-cover`  to `--hide-cover`.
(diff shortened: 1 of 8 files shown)


tramhao:
Sorry that I was on a business trip and no access to github. Is the PR finished?

hasezoey:
> Is the PR finished?

Yes, i would say it is finished, unless you have some things that need to be changed.

PR #484, opened 2025-05-04
Allow keys `add single / add all` on Database `Result` view

### LINKED ISSUE #483: Add all tracks of album
## Describe what you want

In database view, the current workflow for adding all songs of an album is:

1. Enter "Album"
2. Select album
3. Press enter
4. `shift+l` to add all songs

However, I've caught myself trying the following multiple times:

1. Enter "Album"
2. Select album
3. Press `l`  to in my mind: "add selected item to playlist"

This is not critical, but feels like a nice quality of life, especially since I like to add multiple albums to the list and shuffle them

## Do you have already an idea for the implementation?

If `l` is pressed in the `Result` tab, add all songs of the select item to the playlist

hasezoey:
This PR adds support for the `add single` and `add all` keys to work in the `Result` view / section of the Database layout.

Behavior:
- using `add single`(default `L`) will load that selected thing without question (maybe we should ask for anything other than `Album` or `Artist`?)
- using `add all`(default `shift+L`) will first ask via popup if it is actually wanted, if confirmed, will collect all results and send it to the server, which currently can be massively blocking the TUI as the server is kinda slow resolving the tracks and then responding with done.

fixes #483 

@Sporarum could you review if this is what you wanted?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Change(tui): dont depend on `termusicplayback` anymore.
 - Feat(tui): add new cli option `--disable-cover` which actually disables all cover probing and behaves as if no cover features are enabled.
 - Feat(tui): add native theme support. Useful for example for `pywal`.
+- Feat(tui): allow using "add single / add all" keys on Database `Result` view.
 - Feat(server): for rusty backend, add feature `rusty-simd` to enable SIMD instructions for decoding.
 - Feat(server): on rusty backend, do async decoding instead of JIT decoding for music and podcast files. (fixes #191)
 - Fix(tui): set `ueberzug` command to `--silent`.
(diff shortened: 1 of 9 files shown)


Sporarum:
From your description, it seems to match pretty closely

I wasn't able to build locally, the build succeeds:
`Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.56s`
But I couldn't find the executable in `target`:
```
> ls -a target/
.  ..  CACHEDIR.TAG  debug  .rustc_info.json
> ls -a target/debug/
.  ..  build  .cargo-lock  deps  examples  .fingerprint  incremental
```

Sporarum:
If it helps, I ran `make` with the following shell.nix
```nix
{ pkgs ? import <nixpkgs> {} }:

pkgs.mkShell {

  packages = with pkgs; [
    cargo
    rustfmt
    gnumake
    cargo
    rustfmt
    gcc
    pkg-config
    openssl.dev
    alsa-lib.dev
    glib.dev
    gst_all_1.gstreamer
    protobuf
    rustPlatform.bindgenHook
    clippy
 ];

}
```

hasezoey:
> I wasn't able to build locally, the build succeeds:
> Finished dev profile [unoptimized + debuginfo] target(s) in 1.56s
> But I couldn't find the executable in target:

Did you use `make`? If yes, then sorry i am not experienced with makefiles itself, i personally just run `cargo build` directly.

PR #491, opened 2025-05-09
Add Backend Options

hasezoey:
This PR mainly focuses on adding backend related options, in more details:
- add config option to select the backend; previously this was only available as a cli option / environment variable
- add config option to enable / disable soundtouch, if compiled in (otherwise it will have no effect)
- add config options for the other backends (which are skipped until there actual options)

Some slight related refactors:
- move rusty `player_thread` arguments into a struct to reduce confusion and resolve `clippy::too_many_arguments`
- always call `message_on_end` on the sink instead of having to do it outside

I am not quite satisfied with the way i am passing around options from the player_thread to the sink, but without further refactors (which i intend to maybe do) it is ugly, but it will do for now.


hasezoey:
As a additional change, i have refactored the `message_on_end` thing to not require a new thread for each sound, now at the end of a sound a single callback is called which does both everything `rodio::source::Done` and `sleep_until_end` had done.
A custom implementation as rodio (to my knowledge) does not have proper structs for this.

PR #492, opened 2025-05-11
Update dependencies and fix soundtouch-ffi issue

### LINKED ISSUE #485: Compilation gets stuck at soundtouch-ffi(build) step
## Environment & Versions

- System: Linux
- Termusic version: Latest master, from AUR package.
- Rust version: 1.86.0

## Description of the Problem

Compilation have now been ongoing on the usual `termusic-git` package from AUR (https://aur.archlinux.org/packages/termusic-git) for almost 12H for the second time. It's now stuck for hours and hours and hours on this steps:

`    Building [=======================> ] 666/671: soundtouch-ffi(build)`

The remaining thread uses CPU.

Last good build was `termusic-git-0.10.0.r179.g9483e68-1-x86_64.pkg.tar.zst`, I unfortunately can't match it with git to a commit here (I'm not sure how the ID is "crafted" here), but it's probably the last commit from more than 48H. Build was already having this issue yesterday and still have today on this second try.

Did something change so that there are more crates to build compared to before, and that it gets stuck on this one for so long?

hasezoey:
This PR routinely updates some dependencies, most importantly it fixes the `soundtouch-ffi` issue.

Also played around a bit with the lockfile to drop some extra dependencies

fixes #485


hasezoey:
@tramhao @Porkepix could you test to confirm this fixes the soundtouch compilation issue?

hasezoey:
> Two questions I'm curious about though:
> The commented line on https://github.com/Cyanistic/soundtouch-ffi/pull/1/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42R46
> `// let out_dir = std::env::var("OUT_DIR").expect("OUT_DIR is not set by cargo");`

You mean this? I commented it out as rust (and clippy) were telling me that variable was unused, and in that same PR i tried to resolve all the warnings.

> And would you know why https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=termusic-git always ends-up building termusic twice?

From what i can tell, it builds termusic once in *release mode*(in `build()`; `cargo build --release`) and also builds in in *debug-test* mode in `check()`(`cargo test`). I do not know why it is wanting to run the full test suite, but that is up to the package i assume.
The different to `bottom-git` is is that it builds & tests in release mode (`cargo build --release` & `cargo test --release`)

So my best suggestion is adding `--release` to the `cargo test` command, which *should* reduce it somewhat.

hasezoey:
> Wouldn't it be better in that case to just get rid of it rather than comment dead code? I'm guessing it's unlikely to get used in the future as the repo don't have much attention!

For this specific case, i think commenting-out was the quickest option, it might still be necessary and the file also contained a lot of extra commented-code.

> Despite using Manjaro and looking at the work you do on termusic, I guess you don't use at all this kind of -git package as it's easier for you to use your work builds and add the PATH from there?

I do use AUR things, but i dont have any `-git` aur packages installed, because for rust specific things, i just like to compile them myself on a per-user basis (like `bottom`), and also because i cross-compile some (like `bottom`) for my slow raspberry pi.

hasezoey:
> What do you mean by "per-user basis"? -git packages can be for non-rust things, but I gotta admit most of the ones I use are indeed rust ones :p

AUR (via pacman) installs system-wide, for something things i compile myself i like to have them just be in the user's specific PATH (sometimes i also just link the binary into `/usr/bin/` if i am lazy though).

> And moving EDIT from above as I wrote it at the same time as your answer:
>Ah, and I thought they might have wanted to update build against latest soundtouch (2.4.0), as it seems some things are hardcoded against 2.3.2 looking at [Cyanistic/soundtouch-ffi@main/soundtouch-2_3_2](https://github.com/Cyanistic/soundtouch-ffi/tree/main/soundtouch-2_3_2)

I had seen that before, but had no comment on it, as my PR was just about fixing the compilation issue, from what i can tell, soundtouch 2.3.2 is completely fine with the purpose we use it for. Maybe open another issue requesting a upgrade?

hasezoey:
> Single-user here so I don't pay attention to this while feeling it very convenient to be able to update this from the general packages manager

Also a single user here, it is just easier to have non-AUR for cross-compiling and live development(modify code, compile, run without path) while then also being able to lazily use it globally. The rest where this does not apply (at least for rust) is just because i am used to it from the other packages.

> Thing is, I don't know enough about the topic to know how does it matter, if that's important or the consequences (I guess nothing broke on new version, but maybe some new things might be available, missing in this case, but I absolutely don't know).

Will just open a new issue, lets see where this goes.

hasezoey:
@Porkepix 

Re:

> And would you know why https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=termusic-git always ends-up building termusic twice?

> From what i can tell, it builds termusic once in release mode in build() (cargo build --release) and also builds it in debug-test mode in check()(cargo test)
> So my best suggestion is adding --release to the cargo test command, which should reduce it somewhat.

I have tried to add this change to `termusic` arch package, but it was rejected due to running tests in release mode being against arch's rust packaging guidelines:

> Avoid using the --release flag when running tests, because it enables compiler optimizations and disables some features like integer overflow checks and debug_assert!() macro, so in theory you could end up catching less problems.

[Rust package guidelines: Check](https://wiki.archlinux.org/title/Rust_package_guidelines#Check)

[rejected MR](https://gitlab.archlinux.org/archlinux/packaging/packages/termusic/-/merge_requests/1#note_282427)

Though i dont know if this applies to AUR packages, so it might be possible for `termusic-git`.

Issue #493, opened 2025-05-11
Potential slow open with large `playlist.log` and slow backing storage


hasezoey:
## Environment & Versions

- System: Manjaro 25.0.1
- Termusic version: 4db5d006c11c2658b81492619e06fcbc525129eb

- Rust version: 1.86.0

<details>
<summary>Logs</summary>

```txt
[2025-05-11T12:52:41.040+02:00 TRACE termusic_server]: Backend from CLI
[2025-05-11T12:52:52.989+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.056+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.144+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.204+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.312+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.440+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.463+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.485+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.551+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.564+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.655+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.677+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.765+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[2025-05-11T12:52:53.805+02:00 WARN  lofty::mpeg::properties]: MPEG: Using bitrate to estimate duration
[...]

PR #497, opened 2025-05-14
`lib::track` rewrite

hasezoey:
This PR completely re-writes `lib::track` module to be stricter about types and to not store unnecessary stuff all the time.
Some other changes that come along with this:
- extra string allocations are now avoided for Duration formatting (via Display impl)
- add a completely configureable way to parse specific metadata (`lib::track::parse_metadata_from_file`)
- add new `tui::..::tag_editor::te_track` module with Tag-Editor specific track data, which now handles saving tags
- pictures are now loaded on-demand and are globally cached (except for tag-editor, which always fetches freshly from disk)
- lyrics are now loaded on-demand and are globally cached (except for tag-editor, which always fetches freshly from disk)
- reduce some playlist code duplication regarding checking that the grpc data is referencing the correct data

Due to how Pictures are now handled, this PR also fixes the accidental issue i had introduced in #495 that pictures were not loaded at all on any track that gets loaded at initialization (but was working if added via grpc).

For example, Pictures & lyrics are now parsed on-demand and only stored up to 5 at a time, instead of everything being loaded all the time.

This has some interesting data, my main (somewhat) surprise was that the TUI's peak(and average) memory consumption dropped from 246MB(before the PR) to just 45MB(with the PR). Some more metrics:

||before|after|
|:---:|:---:|:---:|
|server peak ram|9.0MB|8.8MB|
|tui peak ram|246MB|45MB|
|server startup time|94s|85s|

With the test load from #493 (and caches dropped).
(ram data collected via [heaptrack](https://github.com/KDE/heaptrack); time data by comparing log times from before and after loading)
(in `bottom` terms, the tui dropped from `1.7%` ram usage to `0.5%`)
Note: Due the earlier mentioned issue introduced in #495, the server did no load any picture data at initialization, so the memory usage basically stays the same before/after the PR; it is safe to assume when comparing to before #495 and after this PR, the RAM usage for the server would be similar to the TUI like in the above table.

re #493.

This PR is marked DRAFT for some extra testing & feedback before being accidentally merged early.


tramhao:
Seems very promising:)

hasezoey:
Updated to include the rust 1.87 lints fixes, also reworded the changelog entry.

I think this PR is ready now.

Weirdly, i just encountered a issue where the what plays and what the server's playlist think its playing de-syncs, though i cant actually reproduce it (happens randomly?) and i dont think its related to this PR.

Issue #159, opened 2023-10-16
TUI should never block to do some async / long task


hasezoey:
Currently when i accidentally press "Backspace" again while i am in the music directory, it will try to search a way wider directory area and the process cannot be stopped with CTRL+C (or thereof) and needs to be killed (kill -9) because it takes so long.

in my opinion the UI should still respond to *at least* stop requests, maybe extend this to allow just stopping the search and go back the the last known path.

PS: it should also have some kind of indicator that it is doing something like a loading bar or some dialog

hasezoey:
A *little* work has gone into this with #443.

hasezoey:
Update: #503 basically fixes the main issue described in this issue, but will not close this issue as, to my knowledge, there are still some other sync things that could block processing commands.

PR #503, opened 2025-05-28
Change the TUI Music library tree loading to be async

### LINKED ISSUE #481: Very slow to open with large music directory
## Environment & Versions

- System:  Linux NixOS
- Termusic version: 0.10.0

Logs:
no relevant logs, no errors.

## Description of the Problem

termusic takes ~5 minutes to load the interface when started with my large music library directory (~900gb flacs) 

it takes this long every time it starts, is there anything that can be done to improve the load speed?

hasezoey:
This PR changes the TUI music library tree loading to be non-blocking and feel responsive by not being locked-up.
Also fixes a bug that the root node is always selected when stepping out of the current directory instead of focusing the stepped-out of node.

Important notes:
- because it is now loading async and the component itself was not touched, means that while something is loading, the user can still navigate and execute things on the tree, like for example starting another load
- the initial `Loading...` text is currently handled via a tree with such a node and a dummy path, but because of this, it can still be navigated. (at least on linux, didnt test this dummy path on windows)
- this pr did not change anything regarding pre-loading a directory, so when encountering a directory that is not pre-loaded (due to for example having `library_scan_depth` set to `1`), the node cannot be expanded (via `ARROW RIGHT`), but can still be stepped into.

@emenel Could you try this PR on your big music library with a higher `library_scan_depth` again and see if it is responsive now?

fixes #481
re #159; this PR coves the main case described there, but does not entirely fix it as, to my knowledge, there are still some sync things that could block processing command.

PS: the previous implementation re-loaded the tree from disk quite often, now it does not anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Fix(tui): re-select the (approximately) same playlist item after a shuffle.
 - Fix(tui): escape key will now no longer also act as a quit key.
 - Fix(tui): changed the music library tree loading to be async.
+- Fix(tui): in the music library tree, when stepping out now the correct node if focused instead of the root node.
 - Fix(server): with `rusty-soundtouch` on `rusty` backend, dont take initial samples until necessary.
 - Fix(server): on rusty backend, always decode and use `f32` samples. (instead of `i16`)
 - Fix(server): on rusty backend, update `soundtouch` version to fix build issues on latest arch & gcc 15.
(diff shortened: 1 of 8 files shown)


emenel:
Thanks! I will give this a try as soon as I have a moment.

emenel:
Hi! Finally trying the new build (from HEAD after this merge).

The UI now loads nice and fast. But the library continues to be empty until the entire scan is done, which is still taking a very long time (over 5 minutes...)

PR #511, opened 2025-06-07
Library Database v2

hasezoey:
This PR implements a new database for storing tracks and related metadata as was discussed in #291 (with some adjustments).
For now this new database is a literal replacement from the old tracks database and does not merge podcasts into it yet.

Note that the old database file is not deleted automatically, similar to how the old config file was not auto deleted.

Likely interesting notes:
- a new server config section was added called `metadata` for metadata parsing & scanning.
- the old `library_scan_depth` is soft-removed (as in it is still de-serialized, but not serialized again).
- for the TUI music tree, what was previously `library_scan_depth` is now a static `2` until it will be refactored to fetch new trees on leaf open.
- because the new database's entries are unique, the bug of termusic repeatedly inserting the same path has been fixed.

Some things i want to do once this is fully integrated:
- possibly merge the podcast database in too
- add a `favorites` feature for tracks, artists and albums
- move the database operations to be on the server instead of the TUI
- add more search options like `Recently added in X days`
- add cleanup / maintenance function (for example remove albums or artists which arent linked anywhere; or tracks which dont exist anymore)

This PR is a DRAFT for some further testing, i would like to hear back if there are any errors or warnings with the new scanning that are *not* `Unknown Formart`. (see TUI logs)

Tagging some people who might be interested or could provide feedback with their music library: @tramhao @Porkepix @Refael10ru


Refael10ru:
Will compile this branch and see tommorow

hasezoey:
Update: rebased on latest master and changed some sql names to be more consistent. (note that the sql / database is not considered stable until this PR is merged)

hasezoey:
Updated to rebase on latest master, the changes include:
- remove already included commits
- remove the custom `Either` implementation
- fix formatting for the new rust edition

hasezoey:
It looks stable enough to me now to be merged where the rest can be ironed out. It is at least better than the old database storage.

PR #518, opened 2025-06-18
Deduplicate `database` widget code & some other things from #511

hasezoey:
This PR moves some things i had done in #511 to a separate PR to include it for the next version and to reduce the review size of #511. In more details what this includes:
- add album artist parsing to the track metadata as a option
- add split album artists and track artists to track metadata as a option
- refactor TUI `database` file to de-duplicate common code paths
- refactor some message types to use better types instead of casting everywhere (`DBMsg::SearchResult`, `PLMsg::Add`)
- refactor some `lib::utils` functions to directly take paths and reduce "if" guards
- make use of `indoc` to have consistent multi-line strings, without having those extra spaces in the compile
- limit vertical size of `DBListCriteria` to the number of table items (now it wont grow beyond that size and leave unused space)

UPDATE:
- reduce code size / deduplicate code by using built-in functions on `Layout`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Fix(tui): escape key will now no longer also act as a quit key.
 - Fix(tui): changed the music library tree loading to be async.
 - Fix(tui): in the music library tree, when stepping out now the correct node if focused instead of the root node.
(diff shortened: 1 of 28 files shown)


tramhao:
Should this PR be included in the current version?

PR #525, opened 2025-06-29
Add dlna server capabilities [Draft]

Slesa:
So far, this adds only a new view with key '4' the same way as library does. The idea is to select mp3 from a media server via DLNA.

My question is if this would be the right way, or should the collection be added to the database? Maybe refresh would be more complicated then?


tramhao:
As the sources are quite different in 1/2/3, I think your idea is the right way to go. Thanks!

lib/src/dlna/discovery.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::{error::Error, time::Duration};
+use futures_util::StreamExt;
+use crate::dlna::models::DlnaDevice;
+
+const EXTRA: &[&str; 1] = &["roomName"];
+
+pub  async fn discover_devices() -> Result<Vec<DlnaDevice>, Box<dyn Error>> {
+    let mut dlna_devices = Vec::new();
+    
+    let search_target = URN::device("schema-upnp-org", "MediaServer", 1).into();
```
hasezoey, line 11:
```suggestion
    let search_target = URN::device("schemas-upnp-org", "MediaServer", 1).into();
```

From my basic knowledge of DLNA, this should be `schemas` instead of `schema`.
This way it actually works and finds my jellyfin DLNA server.

hasezoey (changes requested):
Is this PR still a Draft/Work in Progress, or is it really only meant to add the framework for the dlna services? I personally would prefer if either: dont enable and show anything DLNA yet if it is just bare-bones, or add full functionality before merging.
(if it is still a draft/wip, please mark it as such to avoid accidental merging)

Also am i correctly understanding that this PR is aiming to be a DLNA Client/Player?

If my understanding is correct, then i have the following request:
Change the terminology from `dlan*server*` to `dlna*client*` or `dlna*player*` or something along those lines.

Additionally, not really having a specific place in the code:
- run clippy and resolve all the warnings (i would recommend to rebase onto master to resolve the rust 1.88 warnings)
- change result / error types to use either anyhow or a custom thiserror
- some more documentation

PS: i tried to run `termusiclib::dlna::discovery::discover_devices()` to see if it works, but it does not return any devices by default. This is likely due to using `schema-` instead of `schemas-`.
Note that i have a DLNA server running via Jellyfin (plugin) and kodi and `gssdp-device-sniffer` can find it just fine.

> My question is if this would be the right way, or should the collection be added to the database? Maybe refresh would be more complicated then?

Due to the nature of DLNA, i would not recommend putting it into the sqlite database.

Slesa:
> Is this PR still a Draft/Work in Progress, or is it really only meant to add the framework for the dlna services? I personally would prefer if either: dont enable and show anything DLNA yet if it is just bare-bones, or add full functionality before merging. (if it is still a draft/wip, please mark it as such to avoid accidental merging)

If I would have been sure that this is the right way how to implement it, I would have done it on my own, nothing but me and my laptop until its done. But as long as I have so many questions, I thought it would be better to show it to those who have already implemented so much and ask them if it is the right way.

Thank you for all your code review work, I appreciate that. But maybe it was ~12h too early ;-) at least for the discover stuff.

> 
> Also am i correctly understanding that this PR is aiming to be a DLNA Client/Player?
> 

As a first step, I wanted to collect the mp3 collection from the MediaServer and try to play the songs locally. In a second step, I would like to send them to a MediaRenderer, too.


> If my understanding is correct, then i have the following request: Change the terminology from `dlan*server*` to `dlna*client*` or `dlna*player*` or something along those lines.

Probably the dlna_player will be the renderer, won't it? So dlna_client would be best? Or keep the terms of upnp with dlna_media_server and dlna_media_renderer?


> * run clippy and resolve all the warnings (i would recommend to rebase onto master to resolve the rust 1.88 warnings)
> * change result / error types to use either anyhow or a custom thiserror
> * some more documentation

Yes, this can all be done. But please let me fix all the problems first. 


> > My question is if this would be the right way, or should the collection be added to the database? Maybe refresh would be more complicated then?
> 
> Due to the nature of DLNA, i would not recommend putting it into the sqlite database.

And then show everything in the music library tab with all other tracks? Is it possible to play the tracks directly with the http link? Maybe the covers should be supported, too? But a refresh is probably slow then, as every entry has to be checked if it has changed, has been removed or is new. 

Right now, the reading of the complete collection is really fast, as I simply read the All Music container. But I have problems to split it into the tree folder. Would it be better to allow
All Music (with sort)
Albums
Artists
and read the children when opening the tree as all others to?

tui/src/ui/components/dlna_server.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn handle_right_key(&mut self) -> (CmdResult, Option<Msg>) {
+        let current_node = self.component.tree_state().selected().unwrap();
+        let path: &Path = Path::new(current_node);
+        if path.is_dir() {
+            // TODO: try to load the directory if it is not loaded yet.
+            (self.perform(Cmd::Custom(TREE_CMD_OPEN)), None)
+        } else {
+            (
+                CmdResult::None,
+                Some(Msg::Playlist(PLMsg::Add(path.to_path_buf()))),
```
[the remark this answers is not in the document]
Slesa: As a workaround, I thought about a temporary download of the file from the server.

lib/src/dlna/discovery.rs
Slesa: I did not find the typo, so I copied the already running line from my PoC. But thanks.

lib/src/dlna/media_server.rs
```diff
@@ -0,0 +1,120 @@
...
+                    if item_elem.name == "item" {
+                        if let Some(media_item) = self.parse_media_item(item_elem) {
+                            item_count += 1;
+                            println!("{}: {:?} - {} ({:?}) [{:?}] - {}", media_item.id, media_item.artist, media_item.title, media_item.album, media_item.duration, media_item.url);
+                        }
+                    }
+                }
+            }
+        }
+        println!("Found {} elements", item_count);
```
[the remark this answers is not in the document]
Slesa: Yes they are. Sorry

tui/src/ui/model/mod.rs
```diff
@@ -304,6 +311,7 @@ pub struct Model {
 
     pub layout: TermusicLayout,
     pub library: MusicLibraryData,
+    pub media_server: MediaServerData,
```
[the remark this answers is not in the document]
Slesa: I would also recommend to include 'music' in the library field. Names - so hard to find... dlna_media_server then? Not sure if that fits at all. And if it is necessary when writing to the db?

lib/src/config/v2/tui/keys/mod.rs
```diff
@@ -74,6 +74,8 @@ pub struct Keys {
     pub playlist_keys: KeysPlaylist,
     #[serde(rename = "database")]
     pub database_keys: KeysDatabase,
+    //#[serde(rename = "dlnaserver")]
+    //pub dlna_server_keys: KeysDlnaServer,
```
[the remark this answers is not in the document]
Slesa: Not sure which keys will be necessary... server refresh?

lib/Cargo.toml
```diff
@@ -65,6 +66,7 @@ urlencoding.workspace = true #   = "2"
 viuer.workspace = true #   = "0.6"
 walkdir.workspace = true #   = "2"
 wildmatch.workspace = true #   = "2"
+xmltree= ">=0.11"
```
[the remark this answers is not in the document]
Slesa: If a bit honesty is allowed: reading XML with a loop instead of parsing the known paths is horrible

lib/src/config/v2/tui/keys/mod.rs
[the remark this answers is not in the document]
hasezoey: > Not sure which keys will be necessary... server refresh?

You will likely want at least the following keys:
- `load` / `load_all`, similar to the ones already in library (maybe we should also just re-use those keys?)
- `search`, similar to the ones in library (maybe we should also just reuse that?)
- some key to open a popup or something to change the dlna server selection? (or will this be part of the tree?)

And yes `refresh` would also be great. We also need that in the music library, where it currently does not exist (maybe reuse that key then?).

TL;DR: i just asked because it is commented-out code, if it is there for the future when there *are* keys to be added.

tui/src/ui/model/mod.rs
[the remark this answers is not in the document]
hasezoey: > I would also recommend to include 'music' in the library field.

I totally agree there.

> Names - so hard to find... dlna_media_server then? Not sure if that fits at all.

Sure. Worst case, we will just change it later.

lib/src/dlna/models.rs
```diff
@@ -1,26 +1,40 @@
...
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub  struct MediaItem {
+//#[derive(Clone)] //, Serialize, Deserialize)]
+#[repr(C)]
+pub union MediaChild {
+    pub container: ManuallyDrop<Vec<MediaContainer>>,
+    pub children: ManuallyDrop<Vec<MediaItem>>,
+}
```
hasezoey, line 20: Whats the reason for the `repr` and `ManuallyDrop` here?

tui/src/ui/components/dlna_server.rs
```diff
@@ -134,4 +140,152 @@ impl Component<Msg, UserEvent> for DlnaServer {
...
+            }
+            artist_node
+        });
+        for artist in artists {
+            node.children.push(artist);
+        } */
+        let mut songs = 0;
+        for item in container.items {
+            let album = item.album.unwrap_or("Unknown album".to_string());
+            let artist = item.artist.unwrap_or("Unknown artist".to_string());
```
hasezoey, line 279: Consider using the constants from [`lib::library_db::track_db`](https://github.com/tramhao/termusic/blob/ac84456acaec365017dbb52435f5bcc5fb6a2273/lib/src/library_db/track_db.rs#L95) (i know they are in the library currently, but with #511 they would be moved, and they are also currently used elsewhere).

tui/src/ui/components/dlna_server.rs
```diff
@@ -134,4 +140,152 @@ impl Component<Msg, UserEvent> for DlnaServer {
...
+            let element = format!("[{}] {} {} - {}", album, item.track, artist, item.title);
+            let song = Self::mediaserver_item(item.url, element);
+            node.children.push(song);
+            songs += 1;
+            if songs > 1000 {
+                break;
+            }
```
hasezoey, line 286: Is this limiting just for quick testing?

tui/src/ui/model/mod.rs
```diff
@@ -430,6 +431,8 @@ impl Model {
         );
 
         let media_root = Self::loading_tree();
+        Self::discover_servers(tx_to_main.clone()).await;
```
hasezoey, line 434: I would recommend to put this as a tokio task instead of forcing to wait for DLNA before showing the TUI.

tui/src/ui/components/dlna_server.rs
Slesa: Yes, as long a I am not sure how to deal with albums/artists

tui/src/ui/model/mod.rs
Slesa: This is strange... when I use ```tokio::spawn```, I get "future cannot be sent between threads". When I use it exactly as in the songtext with ```tokio::task::block_in_place```, then it does exactly what it says: it blocks., so the starting is delayed, the ui is shown later.

hasezoey:
This has been open for quite a while without any changes and a bunch of conflicts have gathered, so i will close this for now.

Issue #517, opened 2025-06-11
Next version discussion, `0.11.0`


hasezoey:
I think with the amount of changes and time since the last release, it would be time for another release.

As for open PRs, i would like to get #516 in, but #511 should likely wait for the next version.
Also i think with the time since #497 and #429, and some testing(dogfeeding) i did, i think this is ready.
Especially for fixing #485.

Unless someone has some pressing issue like a panic.

tramhao:
I have no objection to releasing a new version. I have problems, though, as currently the invidious search is not stable and do you use this function often? Is it working on your side?

tramhao:
Yes, the search is done by invidious. I'm having problems with this recently. The download is from youtube directly and it works without problems. Maybe it's because my internet connection.

hasezoey:
Will close this after resolving the following as [`v0.11.0`](https://github.com/tramhao/termusic/blob/master/CHANGELOG.md#v0110) is now released.

With this update, it seems like the once deleted tags `0.2.6` and `0.9.1` have come back (re #448), maybe you will need to delete them locally too to not accidentally push them again?

tramhao:
Ok. I'll delete tags 0.2.6 and 0.9.1 from repo and also my local.

hasezoey:
I will close this as `0.11.0` has been released and looks to be stable.

PR #526, opened 2025-06-30
Fix accidental inclusion of debug prints & revert flexi_logger workaround

hasezoey:
This PR upgrades to `flexi_logger` `0.31.2` and reverts the flexi-logger workaround added in #521 due to flexi_logger handling this case properly now.
This also removes a accidental inclusion of a left-over debug print.

I think after this PR a new version is definitely ready (#517).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1492,9 +1492,9 @@ dependencies = [
 
 [[package]]
 name = "flexi_logger"
-version = "0.31.0"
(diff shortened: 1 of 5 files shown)


hasezoey:
> Is this problem urgent?

It is not really, originally i hoped to have it included in 0.11.0 because i forgot to remove a debug print which is visible to users.

PR #539, opened 2025-07-24
Dynamic Layout & Scrolling for the Config Editor

hasezoey:
This PR basically has 2 parts:
- add dynamic Layout(Grid) to (hopefully) improve element accessibility in low width environments
- add Scrolling to (hopefully) improve element accessibility in low height environments

This also means we dont have to worry too much about adding too many elements in a tab or forcing a specific big display size.

This PR depends on #538 and it would be good to include #537 before this PR, as long as those are not merged and this PR not rebased, this will be a DRAFT.

In more detail and additionally this PR does:
- add 2 dynamic grid helpers to automatically calculate the layout and scroll for elements
- de-duplicate a bunch of code for the "Key" tabs
- remove "Blur" messages for the "Key" tabs and instead use a easy to modify array of ids which now control draw order and focus order
- due to those layout changes, config movement is now horizontal first instead of vertical first

@tramhao what do you think of these changes?


hasezoey:
> I tried to use macro but didn't work out. 

Do you mean a new macro i added, or manually writing one?

> It's a huge headache to add new keys.

In any case, now it is not (that much) anymore.

PR #550, opened 2025-08-10
fix(playback::backends::mpv): allow configuring audio-device

hasezoey:
and dont force "ao=pulse" in linux.

This PR is DRAFT until @foxyseta could try it.

re #452

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,13 @@
 - Feat(tui): make the Config editor (at least rudementally) scrollable.
 - Feat(server): add extra config section for metadata scanning and parsing.
 - Feat(server): on rusty backend, allow configuring output sample rate, now defaulting to 48_000Hz (instead of rodio default 44_100Hz).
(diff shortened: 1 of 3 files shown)


foxyseta:
Because the changes seem to be server-side only, is it enough for me to run

```
cargo install --git https://github.com/hasezoey/termusic --branch mpvDevice --features mpv
```

?

hasezoey:
> Because the changes seem to be server-side only, is it enough for me to run

I have not used git install like this before, but i think this should be fine.

As a test i ran:
```sh
cargo install --git https://github.com/hasezoey/termusic --branch mpvDevice --features mpv termusic-server
cargo install --git https://github.com/hasezoey/termusic --branch mpvDevice termusic
```
Both could not be installed at once as the TUI has no feature for the backends, and cargo will complain about it.
If the installed TUI version is not the version of this PR or its base, you will need to install it too (as there are some changes here and there to the communication since the last released version).
TL;DR: it *would* be fine to not compile the TUI, but it is not worth the hassle of having mismatched versions, so install both.

PR #618, opened 2025-11-24
fix(tui): shutdown tokio runtime with timeout

### LINKED ISSUE #603: termusic-server does not close automatically even when the TUI settings are set to exit the server.
## Environment & Versions

- System: Windows 11 (Installed termusic via Scoop)
- Terminal: PowerShell 7 (v 7.5.4 as per screenshots attached)
- Termusic version: Termusic v0.12.0[g] 
- Rust version:  Rust toolchain does not appear to be installed locally (termusic installed via Scoop pre-compiled binary).

Logs:

Log files were **not** generated in the expected location (`%TEMP%` which is `C:\Users\[USERNAME]\AppData\Local\Temp`) even when launching with the `--log-to-file` flag. The initial startup message still *claims* it is logging to `termusic-tui.log` in that directory, but the file is never created or remains empty. Checked `%APPDATA%\termusic` and `D:\ScoopStuff\scoop\persist\termusic` as well, no logs found, with the latter directory not even existing in Scoop's persist directory. (Note that "D:\ScoopStuff\Scoop" really is just the directory I installed scoop in).

## Description of the Problem

When running `termusic` (installed via Scoop) on Windows 11 within PowerShell 7, the application consistently fails to exit cleanly. Quitting the TUI leaves the background `termusic-server` process running, which hangs the PowerShell terminal session indefinitely and requires manual termination.

**Steps to Reproduce:**

1.  Launch `termusic` from a PowerShell 7 terminal.
3.  Wait for the TUI to fully load.
4.  Immediately press 'q' to quit the application. **Note:** This hang occurs even if no music is played or any other interaction takes place.
**Expected Behavior:**

The `termusic` TUI should close, and the `termusic-server` process should terminate cleanly. The PowerShell terminal should immediately return to a usable prompt (e.g., `PS C:\Users\[USERNAME]>`).

**Actual Behavior:**

1.  The TUI closes visually.
2.  The PowerShell terminal prints the last output line: `Server process ID: [PID]` (e.g., `Server process ID: 10508` as shown in the attached screenshot).
3.  The terminal then hangs indefinitely and becomes completely unresponsive.
4.  Crucially, pressing `Ctrl+C` does **not** interrupt the hung process or return control to the user. The orphaned `termusic-server` process appears to ignore the SIGINT signal.

**Workaround:**

The only way to recover the terminal session is to:
1.  Open a *new* PowerShell tab or window.
2.  Manually identify the PID printed in the hung terminal.
3.  Execute `Stop-Process -Id [PID]` (e.g., `Stop-Process -Id 10508`).

Once the orphaned `termusic-server` process is killed from the separate tab, the original terminal becomes responsive again.

**Configuration:**

My config page in the TUI sets "Stop Sever on TUI exit?" to "Yes"

**Logging Issue:**

Despite the application printing `Logging to file "C:\Users\[USERNAME]\AppData\Local\Temp\termusic-tui.log"` on startup, **no log files (`termusic-tui.log` or `termusic-server.log`) are actually generated** in the `%TEMP%` directory. This happens even when explicitly launching with the `--log-to-file` flag. The log files are not found in other potential locations (`%APPDATA%\termusic`, `D:\ScoopStuff\Scoop\persist\termusic`) either. The failure to generate logs might be related to the exit issue.

**Attachments:**
Note: My username is deliberately omitted. 

* Screenshot about to exit termusic via `q` and selecting "Yes" after being prompted "Are you sure you want to quit?"
Image

* Screenshot of the hung PowerShell terminal after quitting the TUI.
PowerShell 7 stuck after quitting TUI

* Screenshot of relevant TUI config (set "Stop Sever on TUI exit?" to "Yes").
termusic TUI config

* Screenshots of the workaround process (manually stopping the server process via Stop-Process
Image
Image

* Proof for no logs being produced (note that the server process ID is different here due to me launching termusic again)

Image

hasezoey:
In addition with #617, fully fixes #603 (or at least works-around) the tui process not exiting if it spawned the server.

TL;DR: tokio (or the windows API) as a bug or annoyance that makes a child stdio task infinitely blocking, and the `tokio::main` macro uses the default shutdown of "infinitely waiting until all tasks have stopped", causing this to be a problem.

re https://github.com/tokio-rs/console/discussions/651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fix(tui): change the progressbar to use less-than-full blocks for smoother display.
 - Fix(tui): change Database `Playlist` search to work on all music roots, instead of current library root.
 - Fix(tui): change tokio runtime shutdown to have a timeout. This fixes infinitely stuck process on windows.
+- Fix(tui): correctly kill the server on windows systems if config option `behavior.quit_server_on_exit` is enabled.
 - Fix(server): on rusty backend, update symphonia to fix various issues.
 - Fix(server): on rusty backend, update `rusty-libopus` to dynamic link without custom environment variable.
 - Fix(server): dont panic if backend in config is unavailable when overridden by cli argument.
--- a/README.md
+++ b/README.md
@@ -71,19 +71,41 @@ Note that using non-default features might increase the MSRV.
 
 #### Dependencies
 
+##### Linux
+
 | Package name (ubuntu) | Package name (arch) | Required | Build-time-only |       Feature       |                      Description                      |   MSRV   |
 | :-------------------: | :-----------------: | :------: | :-------------: | :-----------------: | :---------------------------------------------------: | :------: |
 |         `git`         |        `git`        |    X     |        X        |                     |                    version control                    |          |
 |        `clang`        |       `clang`       |    X     |        X        |                     |       General Build tools (and sqlite compile)        |          |
 |  `protobuf-compiler`  |     `protobuf`      |    X     |        X        |                     | communication protocol between server and client(tui) |          |
 |    `libdbus-1-dev`    |       `dbus`        |    X     |     unknown     |                     |                  MPRIS media control                  |          |
-|   `libasound2-dev`    |        `alsa-lib`         |    X     |     unknown     |                     |                     ALSA headers                      |          |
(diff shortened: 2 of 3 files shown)


tramhao (commented):
I'll check if it works.

PR #623, opened 2025-12-01
Update MSRV to 1.88

hasezoey:
This PR updates the MSRV to 1.88 for a upcoming PR.
This is separate as due to the if-let chaining now being available, a bunch of style changes have been made.

@tramhao There has not been a breaking change since 0.12.0, maybe consider releasing a 0.12.1 *before* this pr is merged?

### DIFF
--- a/lib/src/config/v1/key.rs
+++ b/lib/src/config/v1/key.rs
@@ -559,14 +559,14 @@ impl BindingForEvent {
             }
         }
         let str_lower_case = str.to_lowercase();
-        if str_lower_case.len() < 4 {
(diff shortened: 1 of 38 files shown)


hasezoey:
> I just merged PR https://github.com/tramhao/termusic/pull/626, will it affect releasing a new version?

No #626 is just a plain fix, I wouldnt consider that breaking.

So i still stand by that `0.12.1` could be released before *this pr*(#623, or its follow-up #624) is merged, after which i would consider it breaking enough for a `0.13.0`. Depends on how you want to release.

Issue #627, opened 2025-12-07
UI is very hard to read when using light themes


arirei:
## Environment & Versions

- System: Endeavour OS, Arch-based
- Termusic version: v0.12.0[c] from arch repos

## Description of the Problem

Some UI elements are very hard to read when using light themes in a terminal. Screenshot attached.

<img width="1920" height="1047" alt="Image" src="https://github.com/user-attachments/assets/0781a65d-0c00-43d2-9bc9-d6b81dc77327" />

arirei:
Right, it's rose piné dawn in wezterm,

here's my config snippet with the color scheme:

```lua
local custom = {
	colors = {
		dawn = {
			base = "#faf4ed",
			text = "#575279",
			love = "#b4637a",
			gold = "#ea9d34",
			rose = "#d7827e",
			pine = "#286983",
			foam = "#56949f",
			iris = "#907aa9",
		}
	},
}
```

arirei:
The palette is the same, yes! Sorry about the confusion

hasezoey:
I forgot to ask more explicitly earlier and just remembered, what is the termusic theme set to? If you cant tell due to the colors, it is also listed in the config (`tui.toml`).
Also if it is anything other than `Native`, could you try the native theme (via the theme selector)?

arirei:
Right, it was set to Termusic Default. Native is much better. Maybe it should be the default?

<img width="1920" height="1049" alt="Image" src="https://github.com/user-attachments/assets/7b969b5a-7289-4ac2-ac43-82743a527ca3" />

PR #631, opened 2025-12-13
Less `Color::Reset` usage

hasezoey:
This PR is a follow-up to what i said in #629. This means this fixes all the places that i came across and can fix without changing a upstream package (https://github.com/veeso/tui-realm-stdlib/issues/43 / https://github.com/veeso/tui-realm-stdlib/pull/44). This means colors will be better than before, but still not fully consistent until fixed upstream.

~~Depends on #629~~
~~Will very likely conflict with #630~~

re #627


hasezoey:
Due to this PR and #630 modifying the same files, i will wait to rebase & mark this PR ready until #630 is merged.

PR #645, opened 2026-01-27
Allow Multiple HTTP IP protocol versions at the same time

### LINKED ISSUE #402: Error `Error binding address (os error 10013)` IPv6
## Environment & Versions

- System: windows 
- Termusic version: 0.9.1 

- Rust version: rustc 1.83.0

Logs:
termusic-server.exe
```
[2024-12-26T17:20:43.500+01:00 ERROR termusic_server]: Error: Error binding address: [::1]:50101

Caused by:
    An attempt was made to access a socket in a way forbidden by its access permissions. (os error 10013)
Error: Error binding address: [::1]:50101

Caused by:
    An attempt was made to access a socket in a way forbidden by its access permissions. (os error 10013)
<!--
Logs can be enabled with `--log-to-file`, they can then by default be found in your temporary directory
- in linux its /tmp/termusic-{tui,server}.log
- in windows its C:\Users\<username>\AppData\Local\Temp\termusic-{tui,server}.log
Log location can also be manually set via `TM_LOGFILE=/path/to/tui.log` for the TUI or `TMS_LOGFILE=/path/to/server.log` for the Server.
Log-level can be set via RUST_LOG, see the following for all levels: https://docs.rs/flexi_logger/latest/flexi_logger/enum.Level.html
-->
```
termusic.exe
```
[2024-12-26T16:59:35.371+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T16:59:50.307+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T17:00:00.168+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T17:16:50.527+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T17:18:04.848+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T17:18:16.550+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T17:19:25.869+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.
[2024-12-26T17:23:00.038+01:00 ERROR termusic]: Error: Could not connect within 5 timeout.

```

hasezoey:
This PR changes server & tui to allow (manual) dual-stack sockets.
This is done by adding a new server config property: `[com.emulate_dual_stack]` which defaults to `true`, which in turn will try to listen to both ipv4 and ipv6 on the given port, but does not fail if either has error-ed, but not both.

This should fix #402

Tagging some users involved and hope for feedback on if this PR fixes the issue if the config is `protocol: http` and default address of `::1`: @Owlyat @valkyrieglasc @DrVoidest

This should NOT be merged until some testing is done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### next
 
 - Change: updated MSRV to 1.88.
+- Feat: add support for listening (and connecting) to both IPv6 and IPv4 with `http` connection mode.
 - Feat(tui): change default theme to be "Native".
 - Fix(tui): fix that "native" and "termusic default" theme also get auto-selected in config editor, if active.
 - Fix(tui): fix a bunch of places where colors were not applied at all or not correctly applied.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5135,6 +5135,7 @@ dependencies = [
  "colored",
  "ctrlc",
  "flexi_logger",
+ "futures-util",
  "log",
  "parking_lot",
  "serde",
--- a/lib/src/config/v2/server/mod.rs
+++ b/lib/src/config/v2/server/mod.rs
@@ -457,6 +457,10 @@ pub struct ComSettings {
     pub port: u16,
     /// gRPC server interface / address
     pub address: IpAddr,
+    /// Emulate dual-stack sockets.
+    ///
+    /// This will bind to both `IPv4` and `IPv6` if the address is `::0` / `0.0.0.0` or `::1` / `127.0.0.1` and will not fail if one is unavailable.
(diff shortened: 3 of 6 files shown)


hasezoey:
From the flow of #402, it looks like the issue was that a too-high port number was used, which has been fixed since.
Also for unix sytems, Unix domain sockets (UDS) are a alternative (and the default), so instead of keeping this PR open forever, i will close this and maybe bring it back if the issue should persist.

PR #648, opened 2026-01-27
Fix `rusty-libopus` on windows & enable features for builds

hasezoey:
This PR does 2 things:
- fix bundling `rusty-libopus` for windows by always requiring the `bundled` feature on windows
- enable feature `rusty-soundtouch` on all targets (and release builds)
- enable feature `rusty-libopus` on all linux & windows targets (and release builds)
  - on linux this is possible due to us already requiring mpv installed, which also has a dependency on libopus
  - on windows this is possible due to us always bundling libopus
- enable feature `cover-iterm` on macos

EDIT: this is inaccurate now, read further comments.


hasezoey:
Slightly changed the contents of the PR, now it only includes:
- fix bundling `rusty-libopus` for windows by always requiring the `bundled` feature on windows
- enable feature `cover-iterm` on macos
- add note that `rusty-soundtouch` requires extra dependencies

PR #654, opened 2026-03-03
Update Macos Github CI & enable more features

hasezoey:
This PR mainly revolves around cleaning up the macos CI builds, but also enables some features on all platforms.
In more details:
- dont require `libsixel` or `sound-touch` to be installed in macos builds
- enable arm64 builds for macos
- enable `rusty-soundtouch` for all targets
- enable `rusty-simd` for macos (as to my understanding it is supported since 10.10, which is 2014) and has more limited hardware which *should* support it since then

@Porkepix i could not get `libsixel` to work, it installs fine but upon linking it says it cannot find library `sixel`, maybe you could try locally?

re https://github.com/tramhao/termusic/pull/652#issuecomment-3990282561

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,11 +95,11 @@ jobs:
           key: ${{ matrix.target }}
 
       - name: Run cargo check
-        run: cargo check --target ${{ matrix.target }} --workspace --features cover-viuer-iterm,rusty-soundtouch,rusty-simd
+        run: cargo check --target ${{ matrix.target }} --workspace --features rusty-soundtouch,rusty-simd
         # run: cargo check --workspace --features cover,all-backends
 
       - name: Run cargo test
-        run: cargo test --target ${{ matrix.target }} --workspace --features cover-viuer-iterm,rusty-soundtouch,rusty-simd
+        run: cargo test --target ${{ matrix.target }} --workspace --features rusty-soundtouch,rusty-simd
         # run: cargo test --workspace --features cover,all-backends
 
   test-win:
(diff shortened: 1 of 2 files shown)


Porkepix:
Okay, a possible lead for the `libsixel` situation (but the question of whether `sixels` can be used on macOS still remains, cf. what I mentioned in previous message): 

Homebrew's `libsixel` (https://formulae.brew.sh/formula/libsixel) builds upon https://github.com/saitoha/sixel which if I understand correctly is an old copy of the version version of this lib.

Arch's (https://archlinux.org/packages/extra/x86_64/libsixel/) builds upon https://github.com/libsixel/libsixel a fork, seemingly abandoned, of https://github.com/saitoha/libsixel after maintainer disappeared for years. It seems he came back and slowly caught up on some things, so it question which one should be used as source, but either way I think https://github.com/saitoha/sixel is an old version and one of the others should probably be used. Dunno if that'd solve the issue at hand though.

EDIT: The rust crates landscape about (lib)sixel seems to be quite a mess too.

hasezoey:
> I tested with Wezterm and Alacritty (can also be tested on Linux, code should be the same), and also with iTerm2 and the native macOS terminal (those are mac specific).

So you tested *termusic* on all those terminals, and on all termusic freezes?

> Aside from the warning for an unused variable, the result is just Kitty Support: "local" and then it shows the picture (path is img/ though, not images/).

You tested on the terminal(s) that freeze in termusic? If it displays a picture, it is not actually a direct issue with viuer.

---

If the above is the case, then the problem is *very* weird, the only thing i can think of is that somehow the input that the terminal returns is eaten by something termusic does, and so viuer goes into a infinite loop waiting for the response (and then collecting your key inputs).

PR #657, opened 2026-03-04
Update `viuer` and use `icy_sixel`

hasezoey:
This PR updates viuer to the latest version, which adds support for more terminals.
Also switch to use `icy_sixel` in `viuer` which is a pure-rust (lib)sixel implementation, which makes us no longer require `libsixel` and build on practically any target.

Due to caution, i have not yet enabled the sixel cover feature by default.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,6 @@ jobs:
         run: | 
             brew update
             brew install protobuf
-            # brew install libsixel
             # brew install gstreamer
             # brew install mpv
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,7 +79,6 @@ jobs:
         run: | 
             brew update
             brew install protobuf
-            # brew install libsixel
   
       - name: Install developer package dependencies
         if: matrix.build == 'x86_64-windows'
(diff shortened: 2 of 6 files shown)


Porkepix:
> @Porkepix could you try this PR again? I have reverted to a earlier viuer version for now, which still has `icy_sixel` support, but not the kitty issue. With this version i cannot reproduce the issue in wezterm.

I'd say it looks fine from little testing. I also saw you opened several issues upstream, maybe you should add it's clear regression between 0.10 and 0.11 then, running some bisect on it if you can build from local copy could help upstream.

Is viuer relying on the other create you opened an issue for this, or is it unrelated in the end?

hasezoey:
> maybe you should add it's clear regression between 0.10 and 0.11 then, running some bisect on it if you can build from local copy could help upstream.

I did mention [in viuer](https://github.com/atanunq/viuer/issues/92) ~~that it *is* fine in 0.10.0, but broken in 0.11.0 and~~ the PR that changed it and why practically *everything* is still affected, only that the PR made it more apparent.

EDIT: i re-read what i wrote in the viuer issue, i did not explicitly mention that 0.10.0 is fine, but even then, like i said 0.11.0 only made the issue more apparent.

> Is viuer relying on the other create you opened an issue for this, or is it unrelated in the end?

It is related and unrelated.
The TL;DR is that both tui-realm (via crossterm) and viuer try to read the input (keys). Terminal responses to queries are simply also just "keys" and neither has proper stdin locking (not actually using `std::io::StdIn` for some reason), so it is a race condition to who gets it first (and viuer has a infinite loop to wait for the keys).

PR #664, opened 2026-03-06
Fix general search action keys to follow source-specific key bindings

### LINKED ISSUE #505: Fix usage of `navigation_keys.right` where there should be a `load_track`
## Environment & Versions

- System: any
- Termusic version: b993bee3123440d3f1a95d6a83fb08f21deaac1b
- Rust version: any

Logs: irrelevant

## Description of the Problem

Currently the Search and some other components make use of `navigation_keys.right` where there should be a custom binding or use `library.load_track`.

Should we add a extra `keys` entry section for `search` keys?

SergioChan:
## Summary
- replace general-search popup labels to use source-specific key bindings instead of `navigation_keys.right`
- trigger source actions with their dedicated bindings (`library_keys.load_track`, `playlist_keys.play_selected`, `database_keys.add_selected`)
- keep behavior unchanged for actions themselves; only key binding source is corrected

## Testing
- Not run locally: `cargo` toolchain is unavailable in this environment (`cargo: command not found`)
- Manually verified by code inspection that each popup source now displays and matches the same key binding it consumes

## Related
Fixes #505

### DIFF
--- a/tui/src/ui/components/popups/general_search.rs
+++ b/tui/src/ui/components/popups/general_search.rs
@@ -264,15 +264,7 @@ impl Component<Msg, UserEvent> for GSTablePopup {
                 return Some(Msg::GeneralSearch(GSMsg::TableBlur));
             }
 
-            Event::Keyboard(keyevent)
-                if match self.source {
-                    Source::Library(_)
-                    | Source::Episode
-                    | Source::Playlist
-                    | Source::Database
-                    | Source::Podcast => keyevent == keys.library_keys.load_track.get(),
-                } =>
-            {
+            Event::Keyboard(keyevent) if keyevent == keys.library_keys.load_track.get() => {
                 match self.source {
                     Source::Library(_) => {
                         return Some(Msg::GeneralSearch(GSMsg::PopupCloseLibraryAddPlaylist));


tui/src/ui/components/popups/general_search.rs
```diff
@@ -265,7 +264,16 @@ impl Component<Msg, UserEvent> for GSTablePopup {
...
 
-            Event::Keyboard(keyevent) if keyevent == keys.navigation_keys.right.get() => {
+            Event::Keyboard(keyevent)
+                if match self.source {
+                    Source::Library(_) | Source::Episode => {
+                        keyevent == keys.library_keys.load_track.get()
+                    }
+                    Source::Playlist => keyevent == keys.playlist_keys.play_selected.get(),
+                    Source::Database => keyevent == keys.database_keys.add_selected.get(),
+                    Source::Podcast => false,
```
hasezoey: Why is this not handled? Is this a path that is currently not called? Even in that case it should likely fallback to `keys.library_keys.load_track`.

tui/src/ui/components/popups/general_search.rs
```diff
@@ -279,7 +287,9 @@ impl Component<Msg, UserEvent> for GSTablePopup {
                     Source::Episode => {
                         return Some(Msg::GeneralSearch(GSMsg::PopupCloseEpisodeAddPlaylist));
                     }
-                    Source::Podcast => return None,
+                    Source::Podcast => {
+                        return Some(Msg::GeneralSearch(GSMsg::PopupCloseOkPodcastLocate));
+                    }
```
hasezoey: I didnt have a look before, so i guess this codepath really was just unused.

SergioChan: Yep, agreed — that podcast branch was effectively dead for this keypath before; this PR now routes it to the same locate action used by Enter.

PR #677, opened 2026-03-10
Routine Dependency update

hasezoey:
This PR updates a bunch of dependencies again, most notably:
- update reqwest to 0.13, which will now use rustls by default instead of native
- update tui-realm to 3.3.0 and stdlib to 3.1, which fixes a bunch of style inconsistencies
  - update tuirealm-orx-tree to 0.3.0, which allows more configuration regarding style and does not draw `Clear` anymore
- update to lofty 0.23, which implements `TXX:*ARTISTS` tags and automatically splits them
- update rodio to 0.22, which makes use of NonZero types and updated dependencies

Additionally does, most likely as a result of the dependency updates:
- unify tui & playback windows dependency version
- fix a bunch of stuff i overlooked in #675, most notably
  - wrong order of layouts, when not being exactly on the first element specified
  - panic when the save popup opens
- fix the config editor theme preview still applying the old theme in unused areas
- change to use `cast_(un)signed` functions over `as i/u64` casting (in case of accidental type changes)

EDIT: due to the `windows` crate update, some types changed and i cannot find how to easily change those, so i decided to just rip it out and hope it works; someone with windows please test the media controls & popup still work.

### DIFF
--- a/playback/src/mpris.rs
+++ b/playback/src/mpris.rs
@@ -276,15 +276,16 @@ impl GeneralPlayer {
 #[allow(clippy::cast_possible_truncation, unsafe_code)]
 mod windows {
     use std::io::Error;
-    use std::mem;
+    // use std::mem;
 
     use windows::core::w;
     // use windows::core::PCWSTR;
-    use windows::Win32::Foundation::{HWND, LPARAM, LRESULT, WPARAM};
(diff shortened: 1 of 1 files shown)


tramhao (commented):
Tested on Windows 11 and works.

PR #691, opened 2026-04-04
Quit only when last connection is dropped

### LINKED ISSUE #687: Settings were not saved under windows
## Environment & Versions

- System:  windows 11 
- Termusic version: master head as of March 27.

- Rust version: 1.94.1

Logs:

### LINKED ISSUE #690: Build failure - commit 0153a92
## Environment & Versions

- System: ArchLinux
- Termusic version: master with AUR
- Rust version: 1.94.1

Logs:

<details>
<summary>Logs</summary>

```txt
   Compiling zbus_names v2.6.1
   Compiling rand v0.10.0
   Compiling alsa v0.11.0
error: failed to run custom build command for `termusic-lib v0.12.1 (/home/clement/.cache/paru/clone/termusic-git/src/termusic-git/lib)`

Caused by:
  process didn't exit successfully: `/home/clement/.cache/paru/clone/termusic-git/src/termusic-git/target/release/build/termusic-lib-bad9ed1b4d17881e/build-script-build` (exit status: 1)
  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: proto/monitor.proto: No such file or directory\n" }
warning: build failed, waiting for other jobs to finish...
==> ERROR: A failure occurred in build().
    Aborting...
error: failed to build 'termusic-git-0.12.1.r169.ga0e38cc-1':
error: packages failed to build: termusic-git-0.12.1.r169.ga0e38cc-1
```

</details>

## Description of the Problem

Termusic fails to build with the above error. This is caused by commit 0153a92 as previous commit was building without any issue.

~~Commit doesn't seems to have introduced a new dependency (as the error could lead to think it was a missing dependency), so I don't know exactly what can cause this.~~

EDIT: Commit mention a new file at https://github.com/tramhao/termusic/commit/0153a921483d60db32eebe3ba31c70d7bae1882f#diff-af47d30fcdc801cbb7c45b27ad3845b49e0c591cfea26b78c957a4edfa8746d8R3 that should be found within https://github.com/tramhao/termusic/tree/master/lib/proto but doesn't exist.

hasezoey:
This PR is a alternative to the currently broken implementation done in 0153a921483d60db32eebe3ba31c70d7bae1882f.

This PR does:
- remove the `MonitorService` implementation, as this is now no longer necessary
- move all connection related functionality to own files
- add a new implementation based on connection counting
- add sources to `PlayerCmd::Quit` to identify where it came from (also good for logs)

Behavior now:
- as long as there never had been any connection, dont quit based on that
- if there is only one connection and that connection send a quit, quit
- if there are multiple connections and one sends a quit, ignore the quit but track it
  - if tracked, once the last connection drops (regardless if that connection also sends a quit), the server will quit
  - this is great as we can also "ignore + track" quit commands if there is still some longer task running (ex. indexing, downloads, etc)

In the future we can decide if we want to add a config option to quit the server if the last connections drops (maybe with a timeout?).

I know that there can be a *slight* race condition with the current implementation of when the check of "do we quit?" is done and maybe a client connects since then, but i think this is minor enough as the old implementation also had that problem.

fixes #690
fixes #687

---

@Porkepix would you want to test this?


Porkepix:
@hasezoey As I don't know what this is about, do you want me to just try to build?

Also, while I'm at it even if that's a bit offtopic: was there some reason for disabling the Linux arm64 builds from CI (It's commented)? I was considering opening a PR to re-enable them as this is already built for Homebrew for example.

EDIT: Also, it seems the branch have conflicts.

hasezoey:
> @hasezoey As I don't know what this is about, do you want me to just try to build?

Guess that would be enough. Maybe add a test that on quit both the TUI and server behave (exit) as expected.

> Also, while I'm at it even if that's a bit offtopic: was there some reason for disabling the Linux arm64 builds from CI (It's commented)?

From what i know, that is that way since before i entered this project. I dont see a reason on why it wouldnt be possible, though i dont have a device to properly test on.

> EDIT: Also, it seems the branch have conflicts.

Because tramhao still seems to fix in a different way.

---

@tramhao Instead of the approach of using *another* service (which is unnecessary), how about the approach in this PR?

Porkepix:
> > @hasezoey As I don't know what this is about, do you want me to just try to build?
> 
> Guess that would be enough. Maybe add a test that on quit both the TUI and server behave (exit) as expected.

So, this is very weird: it compiles correctly, but the resulting build is absolutely broken. It doesn't play anything and some of the shortcuts doesn't do anything (`ctrl +b/f`, `space` for play/pause), while others works; it makes me think this is the shortcuts related to the play. This is tested with a build from source on my work's mac, maybe with a non-supported way, as not listed in the `README`: `cargo build --locked`

PS: Okay, that's weirder. The above is broken while `cargo build --locked--release` isn't. I just tested also the main repo, this is also the case for the master. So it seems everything related to play is broken for debug builds.

Quit with `q` works, I don't know what should the other way to quit be?

> > Also, while I'm at it even if that's a bit offtopic: was there some reason for disabling the Linux arm64 builds from CI (It's commented)?
> 
> From what i know, that is that way since before i entered this project. I dont see a reason on why it wouldnt be possible, though i dont have a device to properly test on.

CI is also here for this! :)

PS:  This also highlighted two things: built-in help states we can quit with `escape`, but I think this never allowed to quit, as far as I can remember? And I don't see play/pause in the list, although it's `intuitive` to do it on `[space]`, maybe it should be added (and maybe I'm not thinking about other shortcuts?)

hasezoey:
> PS: This also highlighted two things: built-in help states we can quit with escape, but I think this never allowed to quit, as far as I can remember? And I don't see play/pause in the list, although it's intuitive to do it on [space], maybe it should be added (and maybe I'm not thinking about other shortcuts?)

Yeah `Help` needs a overhaul and full testing.

Currently some keybindings are hardcoded in some places, some are not, some have different meanings (ex `esc` vs `q`). Is the tag editor help even listed?

> So, this is very weird: it compiles correctly, but the resulting build is absolutely broken. It doesn't play anything and some of the shortcuts doesn't do anything (ctrl +b/f, space for play/pause), while others works; it makes me think this is the shortcuts related to the play. This is tested with a build from source on my work's mac, maybe with a non-supported way, as not listed in the README: cargo build --locked
> PS: Okay, that's weirder. The above is broken while cargo build --locked--release isn't. I just tested also the main repo, this is also the case for the master.

That sounds very weird, nothing is touching this code in this PR.
Have you already looked into the logs for potential errors? (ex your music being opus, but not having `rusty-libopus` enabled)
Did you already try fully cleaning the `target` and completely rebuilding?
What was the last version you used? Was it maybe before #659 / #661?

If you can still reproduce this, maybe try to do a git bisect and open a new issue?

> Quit with q works, I don't know what should the other way to quit be?

For now, just asking if it behaves as *you* would expect on quit.

tramhao:
Absolutely, I want to do the same, using existing service, just at the beginning I don't want to mess with existing code.

Porkepix:
> > PS: This also highlighted two things: built-in help states we can quit with escape, but I think this never allowed to quit, as far as I can remember? And I don't see play/pause in the list, although it's intuitive to do it on [space], maybe it should be added (and maybe I'm not thinking about other shortcuts?)
> 
> Yeah `Help` needs a overhaul and full testing.
> 
> Currently some keybindings are hardcoded in some places, some are not, some have different meanings (ex `esc` vs `q`). Is the tag editor help even listed?

Can't tell you about it as I never tried to check this kind of features.

> > Quit with q works, I don't know what should the other way to quit be?
> 
> For now, just asking if it behaves as _you_ would expect on quit.

Yes

> > So, this is very weird: it compiles correctly, but the resulting build is absolutely broken. It doesn't play anything and some of the shortcuts doesn't do anything (ctrl +b/f, space for play/pause), while others works; it makes me think this is the shortcuts related to the play. This is tested with a build from source on my work's mac, maybe with a non-supported way, as not listed in the README: cargo build --locked
> > PS: Okay, that's weirder. The above is broken while cargo build --locked--release isn't. I just tested also the main repo, this is also the case for the master.
> 
> That sounds very weird, nothing is touching this code in this PR. Have you already looked into the logs for potential errors? (ex your music being opus, but not having `rusty-libopus` enabled) Did you already try fully cleaning the `target` and completely rebuilding? What was the last version you used? Was it maybe before #659 / #661?
> 
> If you can still reproduce this, maybe try to do a git bisect and open a new issue?

Okay, I think my answer wasn't correctly structured/worded so that it got you confused:

Debug builds (`cargo build --locked` without `--release`) are broken, and yes it was on a clean repo newly cloned in `/tmp`. This is applicable to your branch, but also to `master` . When `--release` is added, the build works as expected. Same condition. binaries aren't even written to the same directory: `debug` directory versus `release` directory.

I don't know if I ever tested builds with this command before and therefore had functional debug builds in the past though. So a `bisect` would basically be from first commit to last one. Are debug builds working for you? (how do you build when you're developing?) If not did you used them in the past?

tramhao:
Now it's very clean compared to old implementation. Great! Thanks@hasezoey.

hasezoey:
> Debug builds (`cargo build --locked` without `--release`) are broken, and yes it was on a clean repo newly cloned in `/tmp`. This is applicable to your branch, but also to `master` . When `--release` is added, the build works as expected. Same condition. binaries aren't even written to the same directory: `debug` directory versus `release` directory.
> 
> I don't know if I ever tested builds with this command before and therefore had functional debug builds in the past though. So a `bisect` would basically be from first commit to last one. Are debug builds working for you? (how do you build when you're developing?) If not did you used them in the past?

This still sounds very weird.
Yes i use debug builds (no `--release`), but i commonly dont use `--locked`, but i just now tried it and it works fine (both compile & runtime).
Note that `--locked` only means that it installs dependencies *exactly* as they are in the `Cargo.lock` and does not try to automatically update to another version (or some version of this).

> Same condition. binaries aren't even written to the same directory: debug directory versus release directory.

I dont quite know what you mean here. Where binaries (or build artifacts) end up is entirely up to the rust compiler (or cargo, i dont remember) and matches the profile you set (`--release` ends up in `targets/release` and without in `target/debug`, matching the profile name).

If you build it and still can reproduce the runtime issues, could you provide the logs?

PR #695, opened 2026-04-04
Re-enable aarch64-linux (Linux arm64) builds

Porkepix:
Trying to re-enable `aarch64-linux` (Linux on arm64 architecture) builds. It's built successfully at Homebrew, so no reason it wouldn't here.

It feels like older attempts were probably done with emulation, native ARM runner were maybe not existing back then so I cleaned some details that seems on-point on the matter - although I'm clearly unsure here

Other remarks:
- I'm not sure what "cross" steps or jobs are, is this related to cross-compilation? If yes, is this still necessary? It feels like it could be even more simplified and cleared for anything related to cross.
- Workflow files have a lot of things commented here and there, maybe can they require some cleaning?
- Parts related to Homebrew, will this still be used in the future? eg. is there plan to someday have a Tap (private repo)? It's now officially added and auto-updated on Homebrew-core.
- I had some hesitation about adding some linting for these files as pre-commits and/or workflows, but as of now they could be quite unhappy about some details and blocks.
- I probably forgot one or two details to mention here.

Let me know what do you think about those remarks.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,9 +60,9 @@ jobs:
 
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
-          submodules: true
+          persist-credentials: false
 
       - name: Install ${{ matrix.rust }} toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -155,9 +155,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
-          submodules: false
+          persist-credentials: false
 
       - uses: actions/download-artifact@v4
         # with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,18 +22,18 @@ jobs:
     strategy:
       fail-fast: false # don't fail other jobs if one fails
       matrix:
-        build: [x86_64-linux,x86_64-macos,x86_64-windows,aarch64-macos] #, aarch64-linux] #, x86_64-win-gnu, win32-msvc
+        build: [x86_64-linux, aarch64-linux, x86_64-macos, x86_64-windows, aarch64-macos] # x86_64-win-gnu, win32-msvc
         include:
         - build: x86_64-linux
           os: ubuntu-latest
           rust: stable
           target: x86_64-unknown-linux-gnu
           cross: false
-        # - build: aarch64-linux
-        #   os: ubuntu-20.04
-        #   rust: stable
-        #   target: aarch64-unknown-linux-gnu
-        #   cross: true
+        - build: aarch64-linux
+          os: ubuntu-24.04-arm
+          rust: stable
+          target: aarch64-unknown-linux-gnu
+          cross: false
         - build: x86_64-macos
           os: macos-latest
           rust: stable
@@ -71,7 +71,7 @@ jobs:
           targets: ${{ matrix.target }}
 
       - name: Install developer package dependencies
-        if: matrix.build == 'x86_64-linux'
+        if: matrix.build == 'x86_64-linux' || matrix.build == 'aarch64-linux'
         run: sudo apt-get update && sudo apt-get install libasound2-dev libdbus-1-dev pkg-config protobuf-compiler libgstreamer1.0-0 libunwind-dev libgstreamer1.0-dev libmpv-dev
 
       - name: Install developer package dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -22,18 +22,18 @@ jobs:
...
-        # - build: aarch64-linux
-        #   os: ubuntu-20.04
-        #   rust: stable
-        #   target: aarch64-unknown-linux-gnu
-        #   cross: true
+        - build: aarch64-linux
+          os: ubuntu-24.04-arm
+          rust: stable
+          target: aarch64-unknown-linux-gnu
+          cross: true
```
hasezoey:
```suggestion
          cross: false
```

This runs a `-arm` runner, so it shouldnt need `cross`

hasezoey (changes requested):
> I'm not sure what "cross" steps or jobs are, is this related to cross-compilation?

Yes, specifically, they are for [`cross`](https://crates.io/crates/cross) tool.

> Workflow files have a lot of things commented here and there, maybe can they require some cleaning?

That is likely because it was taken from a template and only the relevant parts were enabled and others commented-out.

.github/workflows/release.yml
Porkepix: But I feel like all the parts related to `cross` are anyway meaningless now that GitHub provides native ARM runners, wouldn't it?
EDIT: And yeah, that's an oversight, it was like that and I just uncommented the block. But I feel like we could remove everything mentioning `cross`.

hasezoey: Yes, that was likely from a template. Cleanup can come later than this pr.

Porkepix: You mean, you prefer to just fix this here and not deal with getting rid of everything cross-related yet?
You prefer to deal with it, or want that I give it a try but separately?

hasezoey: You can try, even in this PR. Just mean it is not *necessary* for this PR.

EDIT: personally though, i would get rid of the completely separate `release` workflow and instead merge it into one, so that most parts are tested to actually run instead of only on release.

Porkepix: Okay, variable fixed in the previous last commit, and another commit added on top of it to clean everything cross-related. Looks fine as-is.

A couple more questions:
- Dunno what the strip part does
- Is it wanted to force `cover` feature for Linux, not elsewhere?
- Wondering if workflow could be simplified, at least maybe the "Build release binary" steps could be renamed more cleanly, without listing explicitly every architecture in the name, as it also appears in parenthesis due to the matrix?

EDIT: I could also merge the last two commits, or the ones about updating actions if preferred.

hasezoey: > Dunno what the strip part does

> strip - discard symbols and other data from object files

Or said differently, minifies the binaries that are going to be released by stripping debug data.

[`strip` man page summary](https://man7.org/linux/man-pages/man1/strip.1.html)

> Is it wanted to force cover feature for Linux, not elsewhere?

Kinda. On windows there are practically no terminals that support any protocol (maybe mintty?), on macos the only reasonable protocol is iterm2, whereas on linux all the other protocols have a use.
Though maybe some protocols should be enabled for other targets, it is just not a priority.

> Wondering if workflow could be simplified, at least maybe the "Build release binary" steps could be renamed more cleanly, without listing explicitly every architecture in the name

Definitely.

> as it also appears in parenthesis due to the matrix?

I am not quite sure what you mean with this part.

> EDIT: I could also merge the last two commits, or the ones about updating actions if preferred.

Its fine as-is.

hasezoey (approved):

.github/workflows/release.yml
Porkepix: > > strip - discard symbols and other data from object files
> 
> Or said differently, minifies the binaries that are going to be released by stripping debug data.
> 
> [`strip` man page summary](https://man7.org/linux/man-pages/man1/strip.1.html)

But is there things `strip` still have to remove on a build done with `--release`? It looks like it should already do this, doesn't it?

> Kinda. On windows there are practically no terminals that support any protocol (maybe mintty?), on macos the only reasonable protocol is iterm2, whereas on linux all the other protocols have a use. Though maybe some protocols should be enabled for other targets, it is just not a priority.

My main terminal on mac currently is Wezterm, dunno which protocole does it favor though.

> > Wondering if workflow could be simplified, at least maybe the "Build release binary" steps could be renamed more cleanly, without listing explicitly every architecture in the name

> > as it also appears in parenthesis due to the matrix?
> 
> I am not quite sure what you mean with this part.

See for example one of the workflows of this PR, the final visible name is `Test Linux (ubuntu-latest, 1.88)`, but the original name in the file is `Test Linux`: we get the matrix variants added in the parentheses.

As such, here for the release workflow I don't see the point of adding the hardware architecture in the steps names as we'll have both this and the matrix variants appearing. I think name should be simplified.

hasezoey: > But is there things strip still have to remove on a build done with --release? It looks like it should already do this, doesn't it?

I didnt add the step and never actually considered this, but upon looking it up, rust's profile `release` by default does *not* strip anything (at least if i understand that correctly), but has a option to do it within the compile instead of doing it as another step.
[cargo: Profiles: option: strip](https://doc.rust-lang.org/cargo/reference/profiles.html#strip)

Porkepix: Eh, interesting. I thought it could be made default. Guess that people needs to explicitly require this, same as for LTO optimizations.

PR #697, opened 2026-04-05
Upgrade to symphonia `0.6.0`

hasezoey:
This PR updates the dependency `symphonia` to `0.6.0(-alpha)`.

As can be seen in the current version of `0.6.0-alpha.1`, symphonia 0.6.0 is still in alpha, but it seems to "just work", so once all dependencies are released to crates.io to support that version, this PR can be merged.

For as long as not all dependencies are released to crates.io, this PR will remain a DRAFT.


hasezoey:
Due to termusic 0.13.x now having been released, do you want to already merge this PR, even though it has git dependencies (libopus-adapter) and a ALPHA symphonia @tramhao ?
Note that as along as there are git dependencies, we will not be able to release a new version to crates.io, and it is unknown when symphonia fully releases 0.6.0 or if libopus-adapter will a alpha version for it (see https://github.com/aschey/symphonia-adapters/issues/40).

hasezoey:
Symphonia 0.6.0 has been fully released and the adapter update has also been released. This is now ready to merge.

Also contains a fix for a regression from #711 i noticed while updating this PR: "Play Specific" did not work and would skip to the next one.

PR #698, opened 2026-04-11
Improve workflows

hasezoey:
This PR applies some changes with regards to what was discussed in #695.
In detail:
- (explicitly) limit permissions available
- make the `release` workflow easier to parse at a glance
- add options to test the `release` workflow other than actually releasing things
  - `workflow_dispatch`: manually trigger a run (without releasing), to make sure it actually fully builds and is still valid
  - `pull_request`: make sure it is valid and working when a pull request changes it
  - `schedule`: make sure it stays valid and builds correctly (once every week)

Note that i could *not* make a broader `push` test, as `paths` combined with either `tags` or `branches` creates a AND condition instead of a OR. (in this case, only triggers a workflow run if a tag is pushed *and* that tag modified the given files)
Additionally, i could not figure out a good way to merge `build` and `release` yet, but this is still a improvement.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ jobs:
   lint:
     name: Lint (Linux)
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     steps:
       - name: Checkout sources
         uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


hasezoey:
By the way, it is more than a few month again since the last release, and i would say the current state is rather stable, what do you think about releasing another version soon @tramhao ?

tramhao:
Yes, it is time to release. Just come back from a trip and will do it soon.

PR #716, opened 2026-06-23
feat(playback): publish current track to mpris on startup

jenningsloy318:
When the server starts with startup_state != Playing, publish the loaded playlist's current track to MPRIS (and Discord) as Stopped, instead of leaving MPRIS clients with an empty default metadata snapshot.

Some MPRIS clients (e.g. noctalia v5) evict players whose first introspection returns empty metadata and never re-introspect them, so without this they show no now-playing info until the user manually starts playback.

The status is intentionally Stopped, not Paused: when a client sends MPRIS `Play`, `GeneralPlayer::play` routes Stopped to `resume_from_stopped` (loads the file and starts the backend), whereas Paused would route to `PlayerTrait::resume` (only un-pauses an already-loaded backend, producing silent "playback").

Add `Mpris::stop` to set PlaybackStatus=Stopped, correcting the Playing status that `Mpris::set_track` hard-codes.

- playback/src/mpris.rs: add `Mpris::stop`
- playback/src/lib.rs: add `GeneralPlayer::publish_current_track_stopped_to_mpris`
- server/src/server.rs: call it in player_loop's non-Playing startup branch

### DIFF
--- a/playback/src/lib.rs
+++ b/playback/src/lib.rs
@@ -539,6 +539,41 @@ impl GeneralPlayer {
         }
     }
 
+    /// Publish the playlist's current track to MPRIS (and Discord) as Stopped,
+    /// without starting actual playback or loading the backend.
+    ///
+    /// Used when the server starts with `startup_state != Playing` so that MPRIS
+    /// clients (e.g. noctalia, KDE-style media widgets) see real now-playing
+    /// metadata immediately, instead of an empty default snapshot. Some shells
+    /// (noctalia v5) evict players whose first metadata snapshot is empty and
+    /// never re-introspect, so without this they stay broken until the user
+    /// manually plays once.
+    ///
+    /// Crucially, this leaves `RunningStatus` as `Stopped`. That way when a
+    /// client sends MPRIS `Play`, [`Self::play`] dispatches to
+    /// [`Self::resume_from_stopped`] (which loads the file and starts the
+    /// backend), instead of [`PlayerTrait::resume`] (which would only un-pause
+    /// an already-loaded backend, producing silent "playback").
+    pub fn publish_current_track_stopped_to_mpris(&mut self) {
+        let Some(track) = self.playlist.read().get_current_track().cloned() else {
+            return; // empty playlist; nothing to publish
+        };
+
+        // Record the current track for run_info readers (set_track_mpris_discord,
+        // send_track_changed), but DO NOT call set_status — leave it as
+        // RunningStatus::Stopped.
+        self.run_info.write().set_current_track(track);
+        self.set_track_mpris_discord();
+
+        // `Mpris::set_track` hard-codes PlaybackStatus=Playing; correct it to
+        // Stopped because no backend is actually loaded yet.
+        if let Some(ref mut mpris) = self.mpris {
+            mpris.stop();
+        }
+
+        self.send_track_changed();
+    }
+
     /// Try to enqueue the next track to play, so that it can be seamlessly played.
     pub fn enqueue_next_from_playlist(&mut self) {
         let mut playlist = self.playlist.write();
--- a/playback/src/mpris.rs
+++ b/playback/src/mpris.rs
@@ -102,6 +102,9 @@ impl Mpris {
             .set_playback(MediaPlayback::Playing { progress: None })
             .ok();
     }
+    pub fn stop(&mut self) {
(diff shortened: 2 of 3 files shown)


playback/src/mpris.rs
```diff
@@ -102,6 +102,9 @@ impl Mpris {
             .set_playback(MediaPlayback::Playing { progress: None })
             .ok();
     }
+    pub fn stop(&mut self) {
```
hasezoey, line 105:
```suggestion
    }

    pub fn stop(&mut self) {
```

playback/src/lib.rs
```diff
@@ -539,6 +539,41 @@ impl GeneralPlayer {
...
+    /// an already-loaded backend, producing silent "playback").
+    pub fn publish_current_track_stopped_to_mpris(&mut self) {
+        let Some(track) = self.playlist.read().get_current_track().cloned() else {
+            return; // empty playlist; nothing to publish
+        };
+
+        // Record the current track for run_info readers (set_track_mpris_discord,
+        // send_track_changed), but DO NOT call set_status — leave it as
+        // RunningStatus::Stopped.
+        self.run_info.write().set_current_track(track);
```
hasezoey, line 565: This is a bad idea, at least as long as the invariant `RunInfo` documents to be upheld.

hasezoey (changes requested):
I was not aware that some MPRIS displays just evict `Stopped` players or "first-publish-empty-metadata", that looks to be a implementation problem on their side.

If we *do* publish `MediaPlayback::Stopped`, then this should likely also be updated in `Mpris::update_progress`

I do not like that we are now again depending on a `current_track` being set (while in `RunningStatus::Stopped`), which i had explicitly changed to not be the case, i was also trying to make the difference between the 2 states `Paused` and `Stopped` more explicit:
- `Pause`: there *is* a current track; the audio backend is running; there may or may not be a playlist
- `Stopped`: there *is no* current track; the audio backend may or may not be loaded; there may or may not be a playlist.

I do not know how we should properly handle media info like MPRIS (or discord) for the `Stopped` state while there are consumers which just evict it and never look into it again.

jenningsloy318:
@hasezoey — you were right. I dug into the consumer side and the root cause is unambiguously a bug in noctalia, not something
  termusic should be working around. Filed details at noctalia-dev/noctalia#2950 — short version:

  - termusic publishes a perfectly valid Paused player with CanPlay = CanControl = true and Metadata = { mpris:trackid: "/" }. Per
  MPRIS-2 spec, xesam:title/artist/album are explicitly optional.
  - noctalia's hasStrongNowPlayingMetadata() evicts any player that lacks those three fields on the first GetAll response after
  NameOwnerChanged, and never re-introspects on subsequent PropertiesChanged. A dbus-monitor capture confirms termusic emits
  well-formed PropertiesChanged signals every 500ms that noctalia silently ignores once evicted.

  So your read — "that looks to be a implementation problem on their side" — was correct. I'm going to close this PR; the RunInfo
  invariant should stay as you designed it.

  One side observation that might be worth a separate, smaller PR (independent of this whole question) — feel free to wave off if
  it's not interesting:

  - Mpris::set_track hard-codes MediaPlayback::Playing even when the caller is in Paused/Stopped, which means a track-change while
  paused briefly mis-reports state until the next update_progress tick corrects it. The fix would be to take the current
  RunningStatus as an arg (or look it up) and pick the right MediaPlayback variant, mirroring what update_progress already does.
  Mentioning because you raised the dual-state-consistency point and this is the same shape of bug.

  Thanks for the careful review — saved me from upstreaming a workaround for someone else's bug.

hasezoey:
> One side observation that might be worth a separate, smaller PR (independent of this whole question) — feel free to wave off if it's not interesting:
> 
>     * Mpris::set_track hard-codes MediaPlayback::Playing even when the caller is in Paused/Stopped, which means a track-change while
>       paused briefly mis-reports state until the next update_progress tick corrects it. The fix would be to take the current
>       RunningStatus as an arg (or look it up) and pick the right MediaPlayback variant, mirroring what update_progress already does.
>       Mentioning because you raised the dual-state-consistency point and this is the same shape of bug.

This is true and `Mpris::set_track` is missing documentation, but the intention is that it gets called from `GeneralPlayer::start_play` (ie the thing that sets `RunningStatus::Playing`).
Though even then there is already the edge case of `Mpris::set_track` also getting called on config reload (regardless of what `RunningStatus` is set).

PR #719, opened 2026-06-23
MPRIS: remove wrong potential `RunningStatus` set & reduce progress / position flicker on PAUSE & RESUME

hasezoey:
This PR fixes a issue that was pointed out in https://github.com/tramhao/termusic/pull/716#issuecomment-4779083186, namely that in `Mpris::set_track` (called by `GeneralPlayer::start_play` & `GeneralPlayer::reload_config`) it would set `MediaPlayback::Running`, regardless of what state it actually was in `::reload_config`, until it gets fixed via a Tick via `::update_progress`.

Additionally:
- because souvlaki always requires the progress / position when setting `Running` or `Paused`, it is now passed along, resulting in no flicker to `0.0`.
- Properly show `Stopped` state in MPRIS and discord

@jenningsloy318 in case you want to review or see the changes

### DIFF
--- a/playback/src/discord.rs
+++ b/playback/src/discord.rs
@@ -18,8 +18,9 @@ pub struct Rpc {
 
 enum RpcCommand {
     Update(String, String),
-    Pause,
     Resume(i64),
+    Pause,
+    Stop,
 }
 
 impl Default for Rpc {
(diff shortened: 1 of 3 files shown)


jenningsloy318:
thank you for your quick reply, actually i am not expert in rust, i'd use AI to check it, from claude, it look good for this PR.

PR #720, opened 2026-06-24
feat(server): add periodic podcast synchronization

jenningsloy318:
Add server-internal scheduled job that periodically refreshes subscribed podcasts, downloads new episodes (deduplicated by GUID/enclosure URL), and appends them to the play queue via PlaylistAddTrack.

Features:
- New [synchronization] config section: enable, interval (humantime), refresh_on_startup
- Tokio periodic task mirroring start_playlist_save_interval structure with CancellationToken + interval_at + select!
- Per-podcast error isolation (network/parse failures don't abort others)
- Downloads to per-podcast subdirectories under [podcast].download_dir
- Episode filenames prefixed with position number for sort order
- PlaylistAddTrack extended with new_append_single/new_append_vec API

Tests: 79 total (unit + integration with wiremock mock HTTP server)

### DIFF
--- a/lib/src/config/v2/server/mod.rs
+++ b/lib/src/config/v2/server/mod.rs
@@ -1,7 +1,7 @@
 use std::{
     fmt::Display,
     net::{IpAddr, SocketAddr},
-    num::{NonZeroU8, NonZeroU32},
+    num::{NonZeroU32, NonZeroU8},
     path::PathBuf,
 };
 
@@ -10,11 +10,20 @@ use serde::{Deserialize, Serialize};
 use crate::track::MediaTypesSimple;
 use backends::BackendSettings;
 use metadata::MetadataSettings;
+use synchronization::SynchronizationSettings;
 
 pub mod backends;
 /// Extra things necessary for a config file, like wrappers for versioning
 pub mod config_extra;
 pub mod metadata;
+pub mod synchronization;
+
+#[cfg(test)]
+mod synchronization_phase3_tests;
+#[cfg(test)]
+mod synchronization_phase5_tests;
+#[cfg(test)]
+mod synchronization_tests;
 
 pub type MusicDirsOwned = Vec<PathBuf>;
 
@@ -27,6 +36,8 @@ pub struct ServerSettings {
     pub podcast: PodcastSettings,
     pub backends: BackendSettings,
     pub metadata: MetadataSettings,
+    /// Podcast synchronization settings (automatic feed refresh and download).
+    pub synchronization: SynchronizationSettings,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
@@ -518,13 +529,13 @@ mod v1_interop {
     use std::num::TryFromIntError;
 
     use super::{
-        Backend, ComSettings, LoopMode, NonZeroU8, NonZeroU32, PlayerSettings, PodcastSettings,
-        PositionYesNo, PositionYesNoLower, RememberLastPosition, ScanDepth, SeekStep,
-        ServerSettings, backends::BackendSettings,
+        backends::BackendSettings, Backend, ComSettings, LoopMode, NonZeroU32, NonZeroU8,
+        PlayerSettings, PodcastSettings, PositionYesNo, PositionYesNoLower, RememberLastPosition,
+        ScanDepth, SeekStep, ServerSettings,
     };
     use crate::config::{
         v1,
-        v2::server::{StartupState, metadata::MetadataSettings},
+        v2::server::{metadata::MetadataSettings, StartupState},
     };
 
     impl From<v1::Loop> for LoopMode {
@@ -649,6 +660,7 @@ mod v1_interop {
                 podcast: podcast_settings,
                 backends: BackendSettings::default(),
                 metadata: MetadataSettings::default(),
+                synchronization: super::SynchronizationSettings::default(),
             })
         }
     }
--- a/lib/src/config/v2/server/synchronization.rs
+++ b/lib/src/config/v2/server/synchronization.rs
@@ -1,0 +1,133 @@
+use std::time::Duration;
+
+use serde::{Deserialize, Deserializer, Serialize};
+
+/// Settings for the periodic podcast synchronization task.
+///
+/// This is a **top-level** config section in `ServerSettings` (not nested under `podcast`)
+/// because synchronization is a server-level scheduling concern that orchestrates multiple
+/// subsystems (feed fetching, downloading, playlist management). Nesting under `podcast`
+/// would incorrectly imply it is solely a podcast configuration concern (AC-06, SCENARIO-030).
+///
+/// When absent from the config file, all fields use their defaults
+/// due to `#[serde(default)]` on the struct.
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct SynchronizationSettings {
+    /// Whether automatic podcast synchronization is enabled.
+    /// Default: true
+    pub enable: bool,
+
+    /// How often to check all subscribed feeds for new episodes.
+    /// Accepts human-readable duration strings: "1h", "30m", "2h30m".
+    /// Default: "1h" (3600 seconds)
+    #[serde(with = "humantime_serde")]
+    pub interval: Duration,
+
+    /// Whether to run a full sync immediately on server startup
+    /// before entering the periodic cycle.
+    /// Default: true
+    pub refresh_on_startup: bool,
+
+    /// Maximum number of new episodes to download per podcast per sync pass.
+    /// Only the newest N undownloaded episodes are fetched.
+    /// Set to 0 for unlimited (download all missing episodes).
+    /// Default: 5
+    pub max_new_episodes: u32,
+
+    /// Whether to automatically add downloaded episodes to the playlist.
+    /// Default: true (backward compatible — existing users see no behavior change).
+    pub auto_enqueue: bool,
+}
+
+impl Default for SynchronizationSettings {
+    fn default() -> Self {
+        Self {
+            enable: true,
+            interval: Duration::from_secs(3600),
+            refresh_on_startup: true,
+            max_new_episodes: 5,
+            auto_enqueue: true,
+        }
+    }
+}
+
+/// Deserialization helper that supports both flat fields and a `[synchronization]` wrapper.
+/// When parsed within `ServerSettings`, fields arrive flat. When parsed standalone from
+/// a TOML document with a `[synchronization]` section header, fields arrive nested.
+#[derive(Deserialize)]
+#[serde(default)]
+struct SyncSettingsHelper {
+    enable: bool,
+    #[serde(with = "humantime_serde")]
+    interval: Duration,
+    refresh_on_startup: bool,
+    max_new_episodes: u32,
+    auto_enqueue: bool,
+    synchronization: Option<SyncSettingsFlat>,
+}
+
+/// Flat representation used when TOML has a `[synchronization]` section header.
+#[derive(Deserialize)]
+#[serde(default)]
+struct SyncSettingsFlat {
+    enable: bool,
+    #[serde(with = "humantime_serde")]
+    interval: Duration,
+    refresh_on_startup: bool,
+    max_new_episodes: u32,
+    auto_enqueue: bool,
+}
+
+impl Default for SyncSettingsHelper {
+    fn default() -> Self {
+        let d = SynchronizationSettings::default();
+        Self {
+            enable: d.enable,
+            interval: d.interval,
+            refresh_on_startup: d.refresh_on_startup,
+            max_new_episodes: d.max_new_episodes,
+            auto_enqueue: d.auto_enqueue,
+            synchronization: None,
+        }
+    }
+}
+
+impl Default for SyncSettingsFlat {
+    fn default() -> Self {
+        let d = SynchronizationSettings::default();
+        Self {
+            enable: d.enable,
+            interval: d.interval,
+            refresh_on_startup: d.refresh_on_startup,
+            max_new_episodes: d.max_new_episodes,
+            auto_enqueue: d.auto_enqueue,
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for SynchronizationSettings {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let helper = SyncSettingsHelper::deserialize(deserializer)?;
+        if let Some(nested) = helper.synchronization {
+            Ok(Self {
+                enable: nested.enable,
+                interval: nested.interval,
+                refresh_on_startup: nested.refresh_on_startup,
+                max_new_episodes: nested.max_new_episodes,
+                auto_enqueue: nested.auto_enqueue,
+            })
+        } else {
+            Ok(Self {
+                enable: helper.enable,
+                interval: helper.interval,
+                refresh_on_startup: helper.refresh_on_startup,
+                max_new_episodes: helper.max_new_episodes,
+                auto_enqueue: helper.auto_enqueue,
+            })
+        }
+    }
+}
--- a/lib/src/config/v2/server/synchronization_tests.rs
+++ b/lib/src/config/v2/server/synchronization_tests.rs
@@ -1,0 +1,368 @@
+//! Tests for `SynchronizationSettings` configuration struct.
+//!
+//! These tests cover Phase 1 of Server-Side Podcast Synchronization:
+//! - AC-01: Config section existence with correct defaults
+//! - AC-10: Config serialization roundtrip
+//! - SCENARIO-001: Default config when section absent
+//! - SCENARIO-002: Explicit non-default values honored
+//! - SCENARIO-003: Roundtrip preserves all fields
+//! - SCENARIO-004: Invalid duration string rejected
+//!
+//! Labeling convention: These labels map to BDD scenarios and acceptance criteria
+//! defined in the specification for traceability between tests and requirements.
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use indoc::indoc;
+    use pretty_assertions::assert_eq;
+
+    use crate::config::v2::server::synchronization::SynchronizationSettings;
+    use crate::config::v2::server::ServerSettings;
+
+    // =========================================================================
+    // SCENARIO-001 / AC-01: Default synchronization config applied when section absent
+    // =========================================================================
+
+    /// When a TOML config has no `[synchronization]` section, the defaults should apply:
+    /// - enable: true
+    /// - interval: 1 hour (3600 seconds)
+    /// - refresh_on_startup: true
+    #[test]
+    fn default_config_when_synchronization_section_absent() {
+        // A minimal server config TOML without any [synchronization] section
+        let toml_str = indoc! { r#"
+            [com]
+            port = 5101
+
+            [player]
+            volume = 30
+
+            [podcast]
+            max_download_retries = 3
+        "# };
+
+        let settings: ServerSettings =
+            toml::from_str(toml_str).expect("should parse without error");
+
+        assert_eq!(settings.synchronization.enable, true);
+        assert_eq!(settings.synchronization.interval, Duration::from_secs(3600));
+        assert_eq!(settings.synchronization.refresh_on_startup, true);
+    }
+
+    /// The Default impl for SynchronizationSettings should produce the documented defaults.
+    #[test]
+    fn default_impl_produces_correct_values() {
+        let defaults = SynchronizationSettings::default();
+
+        assert_eq!(defaults.enable, true);
+        assert_eq!(defaults.interval, Duration::from_secs(3600));
+        assert_eq!(defaults.refresh_on_startup, true);
+    }
+
+    // =========================================================================
+    // SCENARIO-002 / AC-01: Explicit synchronization configuration honored
+    // =========================================================================
+
+    /// When the config TOML specifies all synchronization fields explicitly with
+    /// non-default values, the deserialized struct should reflect those values.
+    #[test]
+    fn explicit_non_default_values_deserialized_correctly() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            enable = false
+            interval = "30m"
+            refresh_on_startup = false
+        "# };
+
+        let settings: SynchronizationSettings =
+            toml::from_str(toml_str).expect("should parse explicit values");
+
+        assert_eq!(settings.enable, false);
+        assert_eq!(settings.interval, Duration::from_secs(30 * 60));
+        assert_eq!(settings.refresh_on_startup, false);
+    }
+
+    /// Test with a different non-default interval value to prevent hardcoding.
+    #[test]
+    fn explicit_interval_2h30m_deserialized_correctly() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            enable = true
+            interval = "2h30m"
+            refresh_on_startup = true
+        "# };
+
+        let settings: SynchronizationSettings =
+            toml::from_str(toml_str).expect("should parse 2h30m interval");
+
+        assert_eq!(settings.interval, Duration::from_secs(2 * 3600 + 30 * 60));
+    }
+
+    /// Test with a seconds-only interval to verify varied duration formats.
+    #[test]
+    fn explicit_interval_seconds_only() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            interval = "45s"
+        "# };
+
+        let settings: SynchronizationSettings =
+            toml::from_str(toml_str).expect("should parse 45s interval");
+
+        assert_eq!(settings.interval, Duration::from_secs(45));
+    }
+
+    // =========================================================================
+    // SCENARIO-003 / AC-01, AC-10: Configuration roundtrip preserves all fields
+    // =========================================================================
+
+    /// Serializing and then deserializing SynchronizationSettings with non-default
+    /// values should produce identical output.
+    #[test]
+    fn serialization_roundtrip_preserves_all_fields() {
+        let original = SynchronizationSettings {
+            enable: false,
+            interval: Duration::from_secs(1800), // 30 minutes
+            refresh_on_startup: false,
+            max_new_episodes: 5,
+            auto_enqueue: true,
+        };
+
+        let serialized = toml::to_string(&original).expect("should serialize");
+        let deserialized: SynchronizationSettings =
+            toml::from_str(&serialized).expect("should deserialize roundtrip");
+
+        assert_eq!(original, deserialized);
+    }
+
+    /// Roundtrip with default values should also be stable.
+    #[test]
+    fn serialization_roundtrip_default_values() {
+        let original = SynchronizationSettings::default();
+
+        let serialized = toml::to_string(&original).expect("should serialize defaults");
+        let deserialized: SynchronizationSettings =
+            toml::from_str(&serialized).expect("should deserialize roundtrip defaults");
+
+        assert_eq!(original, deserialized);
+    }
+
+    /// Roundtrip with a complex interval value (multi-unit duration).
+    #[test]
+    fn serialization_roundtrip_complex_interval() {
+        let original = SynchronizationSettings {
+            enable: true,
+            interval: Duration::from_secs(5 * 3600 + 15 * 60 + 30), // 5h15m30s
+            refresh_on_startup: true,
+            max_new_episodes: 5,
+            auto_enqueue: true,
+        };
+
+        let serialized = toml::to_string(&original).expect("should serialize complex interval");
+        let deserialized: SynchronizationSettings =
+            toml::from_str(&serialized).expect("should deserialize complex interval roundtrip");
+
+        assert_eq!(original, deserialized);
+    }
+
+    // =========================================================================
+    // SCENARIO-004 / AC-01: Unparseable interval duration string rejected
+    // =========================================================================
+
+    /// An unparseable duration string should produce a deserialization error
+    /// with a message indicating the parsing failure.
+    #[test]
+    fn invalid_duration_string_produces_error() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            enable = true
+            interval = "not_a_duration"
+            refresh_on_startup = true
+        "# };
+
+        let result = toml::from_str::<SynchronizationSettings>(toml_str);
+        let err = result.unwrap_err();
+        let err_msg = err.to_string();
+        assert!(
+            err_msg.contains("expected") || err_msg.contains("duration"),
+            "error should mention parsing failure, got: {err_msg}"
+        );
+    }
+
+    /// An empty string should also be rejected with a descriptive error.
+    #[test]
+    fn empty_duration_string_produces_error() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            interval = ""
+        "# };
+
+        let result = toml::from_str::<SynchronizationSettings>(toml_str);
+        let err = result.unwrap_err();
+        let err_msg = err.to_string();
+        assert!(
+            err_msg.contains("expected") || err_msg.contains("duration"),
+            "error should mention parsing failure, got: {err_msg}"
+        );
+    }
+
+    /// A numeric value without unit should be rejected with a descriptive error.
+    #[test]
+    fn numeric_without_unit_produces_error() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            interval = "3600"
+        "# };
+
+        let result = toml::from_str::<SynchronizationSettings>(toml_str);
+        let err = result.unwrap_err();
+        let err_msg = err.to_string();
+        assert!(
+            err_msg.contains("expected") || err_msg.contains("duration"),
+            "error should mention parsing failure, got: {err_msg}"
+        );
+    }
+
+    // =========================================================================
+    // AC-01: Backward compatibility - ServerSettings parses with synchronization
+    // =========================================================================
+
+    /// Full ServerSettings should parse correctly even when synchronization section
+    /// is present with all fields specified.
+    #[test]
+    fn server_settings_with_explicit_synchronization_section() {
+        let toml_str = indoc! { r#"
+            [com]
+            port = 5101
+
+            [player]
+            volume = 30
+
+            [podcast]
+            max_download_retries = 3
+
+            [synchronization]
+            enable = true
+            interval = "2h"
+            refresh_on_startup = false
+        "# };
+
+        let settings: ServerSettings =
+            toml::from_str(toml_str).expect("should parse full server settings");
+
+        assert_eq!(settings.synchronization.enable, true);
+        assert_eq!(settings.synchronization.interval, Duration::from_secs(7200));
+        assert_eq!(settings.synchronization.refresh_on_startup, false);
+    }
+
+    /// ServerSettings with a completely empty config should use all defaults including
+    /// synchronization defaults.
+    #[test]
+    fn server_settings_empty_config_uses_all_defaults() {
+        let toml_str = "";
+
+        let settings: ServerSettings =
+            toml::from_str(toml_str).expect("should parse empty config with defaults");
+
+        assert_eq!(settings.synchronization.enable, true);
+        assert_eq!(settings.synchronization.interval, Duration::from_secs(3600));
+        assert_eq!(settings.synchronization.refresh_on_startup, true);
+    }
+
+    // =========================================================================
+    // AC-01: Partial synchronization section (some fields missing, use defaults)
+    // =========================================================================
+
+    /// When only `enable` is specified, other fields should use defaults.
+    #[test]
+    fn partial_synchronization_section_uses_defaults_for_missing_fields() {
+        let toml_str = indoc! { r#"
+            [synchronization]
+            enable = false
+        "# };
+
+        let settings: SynchronizationSettings =
+            toml::from_str(toml_str).expect("should parse partial config");
+
+        assert_eq!(settings.enable, false);
+        // interval should default to 1h
+        assert_eq!(settings.interval, Duration::from_secs(3600));
+        // refresh_on_startup should default to true
+        assert_eq!(settings.refresh_on_startup, true);
+    }
+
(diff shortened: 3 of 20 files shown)


lib/src/config/v2/server/mod.rs
```diff
@@ -27,6 +32,8 @@ pub struct ServerSettings {
     pub podcast: PodcastSettings,
     pub backends: BackendSettings,
     pub metadata: MetadataSettings,
+    /// Podcast synchronization settings (automatic feed refresh and download).
+    pub synchronization: SynchronizationSettings,
```
hasezoey, line 40: If that is podcast related, it should be under `podcast`

lib/src/config/v2/server/synchronization.rs
```diff
@@ -0,0 +1,142 @@
...
+pub struct SynchronizationSettings {
+    /// Whether automatic podcast synchronization is enabled.
+    /// Default: true
+    pub enable: bool,
+
+    /// How often to check all subscribed feeds for new episodes.
+    /// Accepts human-readable duration strings: "1h", "30m", "2h30m".
+    /// Default: "1h" (3600 seconds)
+    #[serde(with = "humantime_serde")]
+    pub interval: Duration,
```
hasezoey, line 25: These 2 options could be condensed into one where `interval` is set to `0` to disable it

lib/src/config/v2/server/synchronization.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Whether to run a full sync immediately on server startup
+    /// before entering the periodic cycle.
+    /// Default: true
+    pub refresh_on_startup: bool,
+
+    /// Maximum number of new episodes to download per podcast per sync pass.
+    /// Only the newest N undownloaded episodes are fetched.
+    /// Set to 0 for unlimited (download all missing episodes).
+    /// Default: 5
+    pub max_new_episodes: u32,
```
hasezoey, line 36: This should be able to be disabled. Consider either using a enum.

lib/src/config/v2/server/synchronization.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+impl Default for SynchronizationSettings {
+    fn default() -> Self {
+        Self {
+            enable: true,
+            interval: Duration::from_secs(3600),
```
hasezoey, line 47: A quick comment as to how much time that is in human readable would be good here too (i know it is in the option above too).

lib/src/config/v2/server/synchronization.rs
```diff
@@ -0,0 +1,142 @@
...
+        } else {
+            Ok(Self {
+                enable: raw.enable,
+                interval: raw.interval,
+                refresh_on_startup: raw.refresh_on_startup,
+                max_new_episodes: raw.max_new_episodes,
+            })
+        }
+    }
+}
```
hasezoey: Why is this necessary?

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
+//! Tests for `SynchronizationSettings` configuration struct.
+//!
+//! These tests cover Phase 1 of Server-Side Podcast Synchronization:
+//! - AC-01: Config section existence with correct defaults
+//! - AC-10: Config serialization roundtrip
```
hasezoey, line 5: What does the `AC` stand for?

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
...
+    use crate::config::v2::server::synchronization::SynchronizationSettings;
+
+    // =========================================================================
+    // SCENARIO-001 / AC-01: Default synchronization config applied when section absent
+    // =========================================================================
+
+    /// When a TOML config has no `[synchronization]` section, the defaults should apply:
+    /// - enable: true
+    /// - interval: 1 hour (3600 seconds)
+    /// - refresh_on_startup: true
```
hasezoey, line 31: I am pretty sure the defaults dont need to be repeated here again (and all the other tests)

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
...
+        let toml_str = r#"
+[com]
+port = 5101
+
+[player]
+volume = 30
+
+[podcast]
+max_download_retries = 3
+"#;
```
hasezoey: Please use `indoc` to make this indent friendly.

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
...
+enable = true
+interval = "2h30m"
+refresh_on_startup = true
+"#;
+
+        let settings: SynchronizationSettings =
+            toml::from_str(toml_str).expect("should parse 2h30m interval");
+
+        assert_eq!(settings.interval, Duration::from_secs(2 * 3600 + 30 * 60));
+    }
```
hasezoey, line 101: I am pretty sure these 2 tests could be combined as defaults are already tested above.

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
...
+            refresh_on_startup: true,
+            max_new_episodes: 5,
+        };
+
+        let serialized = toml::to_string(&original).expect("should serialize complex interval");
+        let deserialized: SynchronizationSettings =
+            toml::from_str(&serialized).expect("should deserialize complex interval roundtrip");
+
+        assert_eq!(original, deserialized);
+    }
```
hasezoey, line 168: If it is testing the complex interval, might as well test that the string value to be consistent (ie that it is `5h15m30s` and not `XXXXXs`)

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
...
+        let toml_str = r#"
+[synchronization]
+enable = true
+interval = "not_a_duration"
+refresh_on_startup = true
+"#;
+
+        let result = toml::from_str::<SynchronizationSettings>(toml_str);
+        assert!(
+            result.is_err(),
```
hasezoey: I dont like this, this only tests that the result is a error, not what has failed.
Please test against the actual error (and / or message). (this applies to all similar tests)

lib/src/config/v2/server/synchronization_tests.rs
```diff
@@ -0,0 +1,355 @@
...
+    fn server_settings_empty_config_uses_all_defaults() {
+        let toml_str = "";
+
+        let settings: ServerSettings =
+            toml::from_str(toml_str).expect("should parse empty config with defaults");
+
+        assert_eq!(settings.synchronization.enable, true);
+        assert_eq!(settings.synchronization.interval, Duration::from_secs(3600));
+        assert_eq!(settings.synchronization.refresh_on_startup, true);
+    }
```
hasezoey, line 272: Isnt this literally tested above already? (it is the same just having the section itself omitted)

lib/src/player.rs
```diff
@@ -461,6 +465,24 @@ pub mod playlist_helpers {
...
+        }
+
+        /// Create a request to append multiple tracks at the end of the playlist.
+        #[must_use]
+        pub fn new_append_vec(tracks: Vec<PlaylistTrackSource>) -> Self {
+            Self {
+                at_index: Self::AT_END,
+                tracks,
+            }
+        }
```
hasezoey, line 479: For simpler maintenance, those function should just refer to `new_single` and `new_vec` with the sentinel, instead of defining all over again.
Also the docs should mention it is just a alias.

lib/src/player_playlist_add_track_tests.rs
```diff
@@ -0,0 +1,269 @@
+//! Tests for `PlaylistAddTrack` API extension (Phase 2).
+//!
+//! These tests cover Phase 2 of Server-Side Podcast Synchronization:
+//! - T-09: AT_END constant equals u64::MAX
+//! - T-10: new_append_single constructs correct PlaylistAddTrack
+//! - T-11: new_append_vec constructs correct PlaylistAddTrack
+//! - T-12: Unit tests verifying AT_END value and constructor behavior
+//!
+//! AC References:
+//! - AC-07: Downloaded episodes appended to end of play queue via PlaylistAddTrack
```
hasezoey, line 10: Once again what does `AC` and `T` stand for here?

lib/src/player_playlist_add_track_tests.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    /// new_append_single should preserve the exact track source provided.
+    #[test]
+    fn new_append_single_preserves_path_track() {
+        let path = "/home/user/podcasts/episode_42.mp3".to_string();
+        let track = PlaylistTrackSource::Path(path.clone());
+        let request = PlaylistAddTrack::new_append_single(track);
+
+        assert_eq!(request.tracks[0], PlaylistTrackSource::Path(path));
+    }
```
hasezoey, line 76: This is a personal preference, but this could be condensed into one test.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
+// Podcast synchronization module.
+// Implements the sync pass logic and task lifecycle for periodic podcast feed refresh and download.
```
hasezoey:
```suggestion
//! Podcast synchronization module.
//! Implements the sync pass logic and task lifecycle for periodic podcast feed refresh and download.
```

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    cmd_tx: &PlayerCmdSender,
+    db_path: &Path,
+) -> Result<SyncPassStats> {
+    let mut stats = SyncPassStats {
+        podcasts_checked: 0,
+        podcasts_failed: 0,
+        episodes_downloaded: 0,
+        episodes_enqueued: 0,
+        episodes_failed: 0,
+    };
```
hasezoey: This should be a default impl.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+
+    if podcasts.is_empty() {
+        return Ok(stats);
+    }
+
+    // Read config values needed for this pass
+    let (download_dir, concurrent_downloads_max, max_download_retries, max_new_episodes) = {
+        let config_read = config.read();
+        let podcast_settings = &config_read.settings.podcast;
+        let sync_settings = &config_read.settings.synchronization;
```
hasezoey, line 220: This line could be inlined

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                            Err(err) => {
+                                warn!("Failed to get episodes for podcast {pod_id}: {err:#?}");
+                            }
+                        }
+                    }
```
hasezoey, line 311: This is too much nesting, this should be a different function

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                let mut pod_download_dir = download_dir.clone();
+                                pod_download_dir.push(dir_name);
+
+                                // Filter to undownloaded episodes, limited by max_new_episodes
+                                // Episodes are ordered by pubdate DESC (newest first)
+                                let undownloaded: Vec<(usize, &_)> = episodes
+                                    .iter()
+                                    .enumerate()
+                                    .filter(|(_, ep)| ep.path.is_none())
+                                    .collect();
```
hasezoey: This should also filter out any already played (and then locally deleted) episodes.

hasezoey: The limit could also already be applied here.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                            replacement: "",
+                                        },
+                                    );
+                                    // Check if file already exists on disk by prefix match
+                                    // (download_file appends pubdate suffix, so exact match won't work)
+                                    let existing_file = if sanitized_title.is_empty() {
+                                        None // avoid starts_with("") matching any file
+                                    } else {
+                                        std::fs::read_dir(&pod_download_dir)
+                                            .ok()
```
hasezoey: This read should only happen once before the loop, not every iteration.
Also note that this is a SYNC (BLOCKING) operation, while still being in a ASYNC block.

hasezoey:
```suggestion
```

Both `Result` and `Option` have a `and_then`, no need to convert

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                let dir_name = sanitize_with_options(
+                                    &pod_title,
+                                    Options {
+                                        truncate: true,
+                                        windows: true,
+                                        replacement: "",
+                                    },
+                                );
+                                let mut pod_download_dir = download_dir.clone();
+                                pod_download_dir.push(dir_name);
```
hasezoey: There should already be the function `lib::utils::create_podcast_dir`, though that is SYNC

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                    };
+
+                                    if let Some(file_path) = existing_file {
+                                        // File exists on disk — register in DB and enqueue
+                                        if let Err(err) = db.insert_file(ep.id, &file_path) {
+                                            warn!("Failed to register existing file in DB: {err:#?}");
+                                        } else {
+                                            let track = PlaylistTrackSource::Path(
+                                                file_path.to_string_lossy().to_string(),
+                                            );
```
hasezoey: This is wrong, even if the episode is downloaded, it needs to be `PlaylistTrackSource::PodcastUrl` with the episode url

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                        }
+                                    } else {
+                                        episodes_to_download.push(EpData {
+                                            id: ep.id,
+                                            pod_id: ep.pod_id,
+                                            title: ep_title,
+                                            url: ep.url.clone(),
+                                            pubdate: ep.pubdate,
+                                            file_path: None,
+                                        });
```
hasezoey: Couldnt you just use the `ep` and modify that instead of repeating everything?

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                }
+
+                                if !episodes_to_download.is_empty() {
+                                    if let Err(err) = std::fs::create_dir_all(&pod_download_dir) {
+                                        warn!("Failed to create podcast download dir {}: {err}", pod_download_dir.display());
+                                        continue;
+                                    }
+
+                                    // Download episodes using channel-drain pattern
+                                    let dl_taskpool = TaskPool::new(concurrent_downloads_max);
```
hasezoey: There should only be one taskpool that shares its `concurrent_downloads_max`.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                            | PodcastDLResult::DLFileCreateError(ep_data)
+                                            | PodcastDLResult::DLFileWriteError(ep_data) => {
+                                                warn!(
+                                                    "Episode download failed: {} - {}",
+                                                    ep_data.title, ep_data.url
+                                                );
+                                                stats.episodes_failed += 1;
+                                            }
+                                        }
+                                    }
```
hasezoey: This should be a task in-of-itself instead of blocking the podcast feed update receiver.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                                                    {
+                                                        warn!(
+                                                            "Failed to record download in DB: {err:#?}"
+                                                        );
+                                                    }
+
+                                                    // Enqueue via PlaylistAddTrack
+                                                    let track = PlaylistTrackSource::Path(
+                                                        file_path.to_string_lossy().to_string(),
+                                                    );
```
hasezoey: Once again: This is wrong, even if the episode is downloaded, it needs to be `PlaylistTrackSource::PodcastUrl` with the episode url

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                // but handle gracefully
+                warn!("Unexpected NewData result in sync pass (expected SyncData)");
+                msg_counter += 1;
+                stats.podcasts_checked += 1;
+            }
+            PodcastSyncResult::Error(feed) => {
+                msg_counter += 1;
+                stats.podcasts_checked += 1;
+                stats.podcasts_failed += 1;
+                warn!("Feed fetch failed for: {}", feed.url);
```
hasezoey, line 329:
```suggestion
                warn!("Feed fetch failed for: {}", feed.url);
                msg_counter += 1;
                stats.podcasts_checked += 1;
                stats.podcasts_failed += 1;
```

Consistency

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+///
+/// The task exits cleanly when `cancel_token` is cancelled (server shutdown).
+///
+/// Only call this function when `config.read().settings.synchronization.enable` is true.
+pub fn start_podcast_sync_task(
+    handle: tokio::runtime::Handle,
+    cancel_token: tokio_util::sync::CancellationToken,
+    config: SharedServerSettings,
+    cmd_tx: PlayerCmdSender,
+    db_path: std::path::PathBuf,
```
hasezoey, line 416:
```suggestion
    handle: tokio::runtime::Handle,
    cancel_token: CancellationToken,
    config: SharedServerSettings,
    cmd_tx: PlayerCmdSender,
    db_path: PathBuf,
```

Those should be imported instead of cluttering the arguments

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                        Ok(stats) => info!("Periodic sync complete: {stats:?}"),
+                        Err(err) => error!("Periodic sync failed: {err:#?}"),
+                    }
+                },
+                _ = cancel_token.cancelled() => {
+                    info!("Podcast sync task shutting down");
+                    break;
+                }
+            }
+        }
```
hasezoey, line 464: This could be combined by adjusting the `interval_at` starttime to be immediate.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        };
+        new_shared_server_settings(ServerOverlay {
+            settings,
+            ..Default::default()
+        })
+    }
+
+    fn make_cmd_channel() -> (
+        PlayerCmdSender,
+        tokio::sync::mpsc::UnboundedReceiver<(
```
hasezoey, line 512:
```suggestion
        UnboundedReceiver<(
```

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+            episodes_enqueued: 9,
+            episodes_failed: 1,
+        };
+
+        assert_eq!(stats.podcasts_checked, 5);
+        assert_eq!(stats.podcasts_failed, 1);
+        assert_eq!(stats.episodes_downloaded, 10);
+        assert_eq!(stats.episodes_enqueued, 9);
+        assert_eq!(stats.episodes_failed, 1);
+    }
```
hasezoey, line 542: From what i can tell, this tests basic rust functionality, which is unnecessary.
(it would be understandable if a constructor was used though)

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    /// When there are no subscribed podcasts, sync_once should return Ok with
+    /// all-zero stats and not attempt any downloads.
+    /// AC-04, SCENARIO-021: First sync with no subscribed podcasts.
+    #[tokio::test]
+    async fn sync_once_no_podcasts_returns_ok_with_zero_stats() {
+        let tmp_dir = tempfile::tempdir().expect("create temp dir");
+        let db_path = tmp_dir.path();
+
+        // Create the database (empty -- no podcasts)
+        let _db = Database::new(db_path).expect("create database");
```
hasezoey, line 587: This should likely also be a helper factory

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    #[tokio::test]
+    async fn sync_once_unreachable_feed_increments_failed_continues() {
+        let tmp_dir = tempfile::tempdir().expect("create temp dir");
+        let db_path = tmp_dir.path();
+
+        // Create database and insert a podcast with an unreachable URL
+        let db = Database::new(db_path).expect("create database");
+        let podcast = PodcastNoId {
+            title: "Unreachable Podcast".to_string(),
+            url: "http://192.0.2.1:1/nonexistent_feed.xml".to_string(),
```
hasezoey, line 636: To prevent unwanted calls to outside sources, this should use `localhost`

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+            episodes_downloaded: 2,
+            episodes_enqueued: 2,
+            episodes_failed: 1,
+        };
+
+        // The stats struct allows representing partial success
+        assert!(stats.episodes_downloaded > 0);
+        assert!(stats.episodes_failed > 0);
+        assert_eq!(stats.episodes_downloaded + stats.episodes_failed, 3);
+    }
```
hasezoey, line 891: Once again as no actual function calls (constructors or anything) are called, this is only testing basic rust functionality and is unnecessary.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        // Since the feed is unreachable, no episodes should have been enqueued
+        let stats = result.unwrap();
+        assert_eq!(stats.episodes_enqueued, 0);
+
+        // No commands should be received
+        assert!(
+            rx.try_recv().is_err(),
+            "No commands should be sent when feed is unreachable"
+        );
+    }
```
hasezoey, line 937: This test does nothing (it does NOT send any commands), so the test name is misleading.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    #[test]
+    fn playlist_add_track_for_sync_uses_at_end() {
+        // Verify that the constructor used by sync produces AT_END index
+        let track = PlaylistTrackSource::Path("/podcasts/new_episode.mp3".to_string());
+        let cmd = PlaylistAddTrack::new_append_single(track.clone());
+
+        assert_eq!(cmd.at_index, PlaylistAddTrack::AT_END);
+        assert_eq!(cmd.tracks.len(), 1);
+        assert_eq!(cmd.tracks[0], track);
+    }
```
hasezoey, line 951: This is not using `sync_once` and so the name and description is misleading.
Also this is a duplicate of a test in another file.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    fn enqueue_uses_path_source_for_local_files() {
+        let file_path = "/home/user/podcasts/episode_new.mp3";
+        let track = PlaylistTrackSource::Path(file_path.to_string());
+        let cmd = PlaylistAddTrack::new_append_single(track);
+
+        match &cmd.tracks[0] {
+            PlaylistTrackSource::Path(p) => assert_eq!(p, file_path),
+            _ => panic!("Expected PlaylistTrackSource::Path for downloaded episode"),
+        }
+    }
```
hasezoey: Once again does not use `sync_once` or anything and so practically tests nothing.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        let _db = Database::new(db_path).expect("create database");
+
+        let config = make_test_config(db_path);
+        let (cmd_tx, _rx) = make_cmd_channel();
+
+        let result = sync_once(&config, &cmd_tx, db_path).await;
+
+        // Explicitly type-check the result
+        let _stats: SyncPassStats = result.expect("should return SyncPassStats on success");
+    }
```
hasezoey, line 1002: Wasnt this tested already already?

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+            image_url: None,
+        };
+        db.insert_podcast(&podcast).expect("insert podcast");
+
+        // Mark ep1 as downloaded
+        let podcasts = db.get_podcasts().expect("get podcasts");
+        let episodes = &podcasts[0].episodes;
+        // Find the episode with guid "ep1-has-path" and mark it downloaded
+        for ep in episodes {
+            if ep.guid == "ep1-has-path" {
```
hasezoey, line 1054: This should likely just use a `Iter::find`

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        // Since the feed is unreachable, the actual download won't happen,
+        // but the filtering logic is what we're testing here.
+        let config = make_test_config(db_path);
+        let (cmd_tx, _rx) = make_cmd_channel();
+
+        let result = sync_once(&config, &cmd_tx, db_path).await;
+        assert!(result.is_ok());
+        // Feed failed, so no downloads occurred
+        let stats = result.unwrap();
+        assert_eq!(stats.podcasts_failed, 1);
```
hasezoey, line 1082: This test does a lot of setup, but does not actually get any different result it wants to test

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                concurrent_downloads_max: std::num::NonZeroU8::new(1).unwrap(),
+                download_dir: db_path.to_path_buf(),
+                ..Default::default()
+            },
+            ..Default::default()
+        };
+        let config = new_shared_server_settings(ServerOverlay {
+            settings,
+            ..Default::default()
+        });
```
hasezoey, line 1146: This should just use the existing config test creation facility and only modify the specific value

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        let (cmd_tx, _rx) = make_cmd_channel();
+
+        let result = sync_once(&config, &cmd_tx, db_path).await;
+        assert!(result.is_ok());
+    }
```
hasezoey, line 1178: `sync_once` may read it, but it does not cause anything observable in this test, so it is practically useless

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        // This call validates that start_podcast_sync_task exists with the
+        // expected signature matching the spec (Section 4.1):
+        // fn start_podcast_sync_task(
+        //     handle: Handle,
+        //     cancel_token: CancellationToken,
+        //     config: SharedServerSettings,
+        //     cmd_tx: PlayerCmdSender,
+        //     db_path: PathBuf,
+        // )
+        super::start_podcast_sync_task(handle, cancel_token.clone(), config, cmd_tx, db_path);
```
hasezoey, line 1217:
```suggestion
        start_podcast_sync_task(handle, cancel_token.clone(), config, cmd_tx, db_path);
```

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        let db_path = tmp_dir.path().to_path_buf();
+        let _db = Database::new(&db_path).expect("create database");
+
+        let config = make_test_config(&db_path);
+        let (cmd_tx, _rx) = make_cmd_channel();
+        let cancel_token = CancellationToken::new();
+        let handle = Handle::current();
+
+        // Start the sync task
+        super::start_podcast_sync_task(handle, cancel_token.clone(), config, cmd_tx, db_path);
```
hasezoey, line 1243:
```suggestion
        start_podcast_sync_task(handle, cancel_token.clone(), config, cmd_tx, db_path);
```

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        });
+
+        // Verify the gating condition: when enable is false, the task should
+        // not be spawned. This mirrors the check in actual_main().
+        let should_spawn = config.read().settings.synchronization.enable;
+        assert!(
+            !should_spawn,
+            "Sync task should NOT be spawned when enable is false"
+        );
+    }
```
hasezoey, line 1295: This does not actually test the "non-execute" of the main function, and the rest is already handled in other test files

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        super::start_podcast_sync_task(handle, cancel_token.clone(), config, cmd_tx, db_path);
+
+        // Wait briefly -- no sync should have occurred
+        tokio::time::sleep(Duration::from_millis(200)).await;
+
+        // No PlaylistAddTrack commands should have been sent (no sync ran)
+        assert!(
+            rx.try_recv().is_err(),
+            "No commands should be sent when refresh_on_startup is false and interval hasn't elapsed"
+        );
```
hasezoey, line 1393: This should likely also be in the `= true` case test as it is a empty database

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        // Wait long enough for at least one periodic tick to fire (50ms interval + margin)
+        tokio::time::sleep(Duration::from_millis(150)).await;
+
+        // Cancel and clean up
+        cancel_token.cancel();
+        tokio::time::sleep(Duration::from_millis(50)).await;
+
+        // If we reach here without panic, the periodic timer fired correctly.
+        // The sync_once calls completed (with empty DB, they return quickly).
+    }
```
hasezoey, line 1448: This test does not test anything other tests have not done already as timerdrift is not actually tested for nor that it has executed or not.

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        // - cancel_token: CancellationToken
+        // - config: SharedServerSettings
+        // - cmd_tx: PlayerCmdSender
+        // - db_path: PathBuf
+        super::start_podcast_sync_task(handle, cancel_token.clone(), config, cmd_tx, db_path);
+
+        // Clean up
+        cancel_token.cancel();
+        tokio::time::sleep(Duration::from_millis(50)).await;
+    }
```
hasezoey, line 1530: Wasnt testing the arguments already a test above?

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        tokio::time::sleep(Duration::from_millis(200)).await;
+
+        // Cancel after startup sync should have been attempted
+        cancel_token.cancel();
+        tokio::time::sleep(Duration::from_millis(50)).await;
+
+        // If we reach here, the startup sync was triggered without waiting
+        // for the interval. With refresh_on_startup=true, sync_once is called
+        // immediately on task start, before entering the interval_at loop.
+    }
```
hasezoey, line 1595: Wasnt this literally a test above?

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    fn generate_rss_feed(title: &str, episodes: &[(&str, &str, &str)]) -> String {
+        let mut items = String::new();
+        for (ep_title, guid, url) in episodes {
+            items.push_str(&format!(
+                r#"
+        <item>
+            <title>{ep_title}</title>
+            <guid>{guid}</guid>
+            <enclosure url="{url}" type="audio/mpeg" length="1024"/>
+            <pubDate>Mon, 23 Jun 2025 12:00:00 +0000</pubDate>
```
hasezoey, line 1619: If the date is static, it may cause inconsistent sorting (sorting based on insert order, which is not actually how it would work)

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+    }
+
+    /// Generate minimal audio file content for download mocking.
+    fn fake_audio_content() -> Vec<u8> {
+        // ID3 header + minimal frame to look like an MP3 file
+        let mut content = vec![0x49, 0x44, 0x33]; // "ID3" magic bytes
+        content.extend_from_slice(&[0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]);
+        content.extend_from_slice(&[0xFF; 1024]); // Fake audio data
+        content
+    }
```
hasezoey, line 1644: As this is static content, this could be a static slice too

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+        assert!(result2.is_ok());
+        let stats2 = result2.unwrap();
+        assert_eq!(
+            stats2.episodes_downloaded, 0,
+            "second pass should not re-download"
+        );
+        assert_eq!(
+            stats2.episodes_enqueued, 0,
+            "second pass should not re-enqueue"
+        );
```
hasezoey, line 1905: This should also have a test that `1` podcast has been checked

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+            stats.episodes_enqueued, 1,
+            "only the new episode should be enqueued"
+        );
+
+        // Verify the command
+        let (cmd, _) = cmd_rx.try_recv().expect("should receive one command");
+        match cmd {
+            PlayerCmd::PlaylistAddTrack(add_track) => {
+                assert_eq!(add_track.at_index, PlaylistAddTrack::AT_END);
+                assert_eq!(add_track.tracks.len(), 1);
```
hasezoey, line 2021: It should also test the correct episode is downloaded

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+                    "must use AT_END for auto-start detection"
+                );
+                assert!(
+                    !add_track.tracks.is_empty(),
+                    "must have at least one track to trigger playback"
+                );
+            }
+            other => panic!("Expected PlaylistAddTrack, got: {:?}", other),
+        }
+    }
```
hasezoey, line 2106: This does not actually test anything new and especially not that the player actually starts

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+            PlayerCmd::PlaylistAddTrack(add_track) => {
+                assert_eq!(
+                    add_track.at_index,
+                    PlaylistAddTrack::AT_END,
+                    "sync during playback must append at end"
+                );
+            }
+            other => panic!("Expected PlaylistAddTrack, got: {:?}", other),
+        }
+    }
```
hasezoey, line 2523: This does not test anything new

server/src/podcast_sync.rs
```diff
@@ -0,0 +1,2691 @@
...
+            .and(path("/url_dedup_feed.xml"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .set_body_string(feed_xml)
+                    .insert_header("content-type", "application/rss+xml"),
+            )
+            .mount(&mock_server)
+            .await;
+
+        // Do NOT mock the episode download -- it should never be requested
```
hasezoey, line 2728: It should likely exist, but use a spy to confirm it was never hit

hasezoey (changes requested):
Thanks for trying your hand at this feature, but i think the current design is not quite right.

In my opinion the podcasts themself should store a `last_checked` (or `next_check_at`) date and potentially a `check_interval`.
This information should then be used to fetch to podcast changes.
Additionally, i dont like that the new episodes get automatically added to the playlist, as currently they can (theoretically) appear in any order between any podcast that gets synchronized and if there is a episode that gets skipped (ex. network error) other episodes just still get appended and that there is currently no way to turn this off.

Finally, i think before this feature is added, the whole existing podcast sync should be moved to the server instead of being handled by the TUI.

Side notes:
- the commit message style, specifically the scope, is not *quite* what we use here (see recent commits and [CONTRIBUTING](https://github.com/tramhao/termusic/blob/master/CONTRIBUTING.md) as reference)
- there should only be one global TaskPool for network (or at the very least podcast) tasks
- things i likely forgot already trying to review the massive changes in the past 3 hours

jenningsloy318:
@hasezoey since in your review, i saw there  a huge redesign decision, move the sync from tui to server, can you confirm if we can move on or wait for your decision ?

hasezoey:
> i saw there a huge redesign decision, move the sync from tui to server, can you confirm if we can move on or wait for your decision ?

It was not just that decision, it was also that i would recommend / prefer if we go with a "last_checked" / "next_check_at" in database instead of always directly re-syncing approach.
This approach, without much extra work, would allow reduced synchronizations when for example restarting often, and allow easier crash / early exit recovery by only having to check the overdue ones. Finally, this would also potentially show the user when the last successful sync for that podcast had been.

@tramhao what do you think which approach would be better?

jenningsloy318:
I fixed the bugs, but still left 2 strategic design questions that fundamentally change the architecture

  1. A1 — Per-podcast last_checked/next_check_at: Instead of "sync all podcasts every N minutes", store per-podcast timestamps in the DB.
  Benefits:
    - Skip recently-checked podcasts on restart
    - Crash recovery only re-checks overdue ones
    - User can see when last successful sync happened
    - Different podcasts could have different check intervals
  2. A3 — Move existing TUI sync to server: The reviewer suggests that before adding periodic sync on the server, the manual podcast sync  (currently in TUI) should first be moved to the server side. This would avoid duplicate sync logic in two places.

hasezoey:
I gave the changes a skim, but it is very hard to read due to all the noise and changes being set as "unviewed" even though nothing had changed (due to it being a single force-pushed commit).
If you do go ahead and change the PR (even though i am still mostly against the design), please split it out into multiple commits. (also read [CONTRIBUTING](https://github.com/tramhao/termusic/blob/master/CONTRIBUTING.md#contribuing-code) about commit and PR style)

Also, not many of my review comments (which are unrelated to the design decision) have seemingly not been addressed at all.

PS: i saw that `std::fs::read_dir` has just been wrapped in a `tokio::spawn_block`, but you know there is `(async)tokio::fs::read_dir` right?

jenningsloy318:
I appoligize for this, since I am not a rust developer, I implemented these changes via AI tool, my purpose is to modify the code to fit my own needs that I want to synchronize it in background.
 Creating this  PR that i am not sure if anyone intrested in this feature. If this bother you, you can just leave there then I can close this PR.

Really appreciated this project, since I used it for a long time.

hasezoey:
To prevent accidental merging, i have converted this to a DRAFT

RG-IL:
Is this going somewhere, or should it be closed?

hasezoey:
> Is this going somewhere, or should it be closed?

Considering that the original PR was made mostly by AI and jenningsloy318 does not have enough rust knowledge to fix this (according to their words, if i read that right), i dont think this is going anywhere.

So i will close this now.

PR #723, opened 2026-06-28
feat(tui::theme): add native transparent for background

theeasternfurry:
+ Make termusic transparent background
+ Add Native Opacity option in theme

image

image

### DIFF
--- a/lib/src/config/v2/tui/theme/mod.rs
+++ b/lib/src/config/v2/tui/theme/mod.rs
@@ -203,6 +203,8 @@ pub enum ThemeColor {
     Native,
     /// Set explicit RGB colors
     Hex(ThemeColorHex),
+    /// Transparent background
+    Transparent,
 }
 
 impl ThemeColor {
@@ -220,10 +222,20 @@ impl ThemeColor {
         Self::Native
     }
 
+    /// Create a new instance for transparent
+    #[must_use]
+    pub const fn new_transparent() -> Self {
+        Self::Transparent
+    }
+
     /// Try to parse a hex or "native" string.
     fn from_string(val: &str) -> Result<Self, ThemeColorParseError> {
         if val == "native" {
             return Ok(Self::Native);
+        }
+
+        if val == "transparent" {
+            return Ok(Self::Transparent);
         }
 
         let res = match ThemeColorHex::try_from(val) {
@@ -248,18 +260,18 @@ impl ThemeColor {
         match self {
             ThemeColor::Native => "native".to_string(),
             ThemeColor::Hex(theme_color_hex) => theme_color_hex.to_hex(),
+            ThemeColor::Transparent => "transparent".to_string(),
         }
     }
 
     /// Resolve the current instance to either native coloring (requires `style`) or a rgb color.
     #[must_use]
     pub fn resolve_color(&self, style: ColorTermusic) -> Color {
-        let hex = match self {
-            ThemeColor::Native => return style.into(),
-            ThemeColor::Hex(theme_color_hex) => theme_color_hex,
-        };
-
-        (*hex).into()
+        match self {
+            ThemeColor::Native => style.into(),
+            ThemeColor::Hex(theme_color_hex) => (*theme_color_hex).into(),
+            ThemeColor::Transparent => Color::Reset,
+        }
     }
 }
 
@@ -797,6 +809,8 @@ mod tests {
     use super::ThemeColors;
 
     mod theme_color {
+        use super::super::Color;
+        use super::super::ColorTermusic;
         use super::super::ThemeColor;
 
         #[test]
@@ -823,6 +837,26 @@ mod tests {
         fn should_serialize() {
             assert_eq!(ThemeColor::new_hex(1, 2, 3).to_string(), "#010203");
             assert_eq!(ThemeColor::new_native().to_string(), "native");
+        }
+
+        #[test]
+        fn should_parse_transparent() {
+            assert_eq!(
+                ThemeColor::new_transparent(),
+                ThemeColor::try_from("transparent").unwrap()
+            );
+        }
+
+        #[test]
+        fn should_serialize_transparent() {
+            assert_eq!(ThemeColor::new_transparent().to_string(), "transparent");
+        }
+
+        #[test]
+        fn should_resolve_transparent_to_reset() {
+            let transparent = ThemeColor::new_transparent();
+            let color = transparent.resolve_color(ColorTermusic::Foreground);
+            assert_eq!(color, Color::Reset);
         }
     }
 
--- a/lib/themes/Transparent.yml
+++ b/lib/themes/Transparent.yml
@@ -1,0 +1,26 @@
+colors:
+  name: Transparent
+  primary:
+    background: "transparent"
+    foreground: "native"
+  cursor:
+    text: "native"
+    cursor: "native"
+  normal:
+    black: "native"
+    red: "native"
+    green: "native"
+    yellow: "native"
+    blue: "native"
+    magenta: "native"
+    cyan: "native"
+    white: "native"
+  bright:
+    black: "native"
+    red: "native"
+    green: "native"
+    yellow: "native"
+    blue: "native"
+    magenta: "native"
+    cyan: "native"
+    white: "native"


lib/src/config/v2/tui/theme/mod.rs
```diff
@@ -220,12 +222,22 @@ impl ThemeColor {
...
+        Self::Opacity
+    }
+
     /// Try to parse a hex or "native" string.
     fn from_string(val: &str) -> Result<Self, ThemeColorParseError> {
         if val == "native" {
             return Ok(Self::Native);
         }
 
+        if val == "opacity" {
```
hasezoey: I dont quite like the name, shouldnt it be `opaque` as there is no slider to control the opacity?
Alternatively, `transparent` would also be a option (and should likely be added as a alias at least)

hasezoey (changes requested):
From what i can tell, Transparent / Opaque background is practically just a feature of the Terminal when `(Color::)Reset` background is used and is not standardized (nor do we as the program know it is possible).
Hence i think this should not be added as a fully separate `ThemeColor` (as it likely only is even valid for the background) and would only be kind of a alias for having the style set to `Reset`. (Though i do fully understand that changing all backgrounds in style is a bit cumbersome)
Other options would be to document that `Reset` could be used to get native background, which include Transparent background.

@tramhao what do you think?

lib/src/config/v2/tui/theme/mod.rs
theeasternfurry: > I dont quite like the name, shouldnt it be `opaque` as there is no slider to control the opacity? Alternatively, `transparent` would also be a option (and should likely be added as a alias at least)

I am not good at naming. But you are right. I think `transparent` should fit for this one. I will rename it as soon as possible.

hasezoey (commented):
Reads better to me now, though i am still on the fence about actually adding a dedicated Theme / `ThemeColor` for it.

theeasternfurry:
@hasezoey  I think it's necessary for now, I could have done better solution by combining colors & theme instead in above comment file but it will be very time consuming and out of scope for this PR.

From what I see `self.theme.background` expects a `ThemeColor`. Without `ThemeColor::Transparent`, there's no way for users to set background to transparent in the config file (since `Color::Reset` isn't a ThemeColor).

Maybe next time we should rebuild whole theme & colors?

hasezoey:
> From what I see self.theme.background expects a ThemeColor. Without ThemeColor::Transparent, there's no way for users to set background to transparent in the config file (since Color::Reset isn't a ThemeColor).

I do not understand your argument, you can still set `Color::Reset` as a style in the config editor (or the config file itself) as `ThemeColor::Transparent` is literally just a mapping to `Color::Reset`.

tramhao:
It looks nice, but if it's possible to do with current config, I'd rather see a theme or theme group called transparent. Thanks.

theeasternfurry:
What about now? Does it better? I am not sure if I can understand correctly so I split them to theme file instead of hardcode it.

hasezoey:
> What about now? Does it better? I am not sure if I can understand correctly so I split them to theme file instead of hardcode it.
> bbb48f9d1c637c4ba331513136a194a76e7b37e5

I literally have no clue what this theme is meant to do, this is a literal copy of a `Native` theme (which is handled in-memory and no actual theme file) with seemingly no actual runtime usage of it or transparency.

Personally, i also dont quite understand what tramhao was wanting to say, but i think it meant something along the lines "handle it similar like `Native`", which was done previously in this PR.

Though once again, i dont agree that this should be a Theme / Color at all as this literally is just a alias for `Color::Reset`, which can be set individually in the style already. Additionally, this `Transparent` (alias `Color::Reset`) would only really be relevant / valid for backgrounds, which would theoretically be possible to set for anything else, but would not make much sense.

Example, the following TUI config should apply your transparency effect while being on current MASTER (0bcc1bc4ebb8740e9f75c06bd62d824230f4ef5a), no PR necessary:

<details>
<summary>For length of this comment, this has been collapsed</summary>

```toml
version = "2"
com = "same"

[behavior]
quit_server_on_exit = true
confirm_quit = true

[coverart]
align = "bottom right"
size_scale = 0
hidden = false
protocols = ["iterm2", "ueberzug", "kitty", "sixel"]

[style.library]
foreground_color = "Foreground"
background_color = "Reset"
border_color = "Blue"
highlight_color = "LightYellow"
highlight_symbol = "🦄"

[style.playlist]
foreground_color = "Foreground"
background_color = "Reset"
border_color = "Blue"
highlight_color = "LightYellow"
highlight_symbol = "🚀"
current_track_symbol = "►"
use_loop_mode_symbol = true

[style.lyric]
foreground_color = "Foreground"
background_color = "Reset"
border_color = "Blue"

[style.progress]
foreground_color = "Foreground"
background_color = "Reset"
border_color = "Blue"

[style.important_popup]
foreground_color = "Yellow"
background_color = "Reset"
border_color = "Yellow"

[style.fallback]
foreground_color = "Foreground"
background_color = "Reset"
border_color = "Blue"
highlight_color = "LightYellow"

[theme]
file_name = "3024.light"
name = "3024 (light)"
author = "Chris Kempson"

[theme.primary]
background = "#f7f7f7"
foreground = "#4a4543"

[theme.cursor]
text = "#f7f7f7"
cursor = "#4a4543"

[theme.normal]
black = "#090300"
red = "#db2d20"
green = "#01a252"
yellow = "#fded02"
blue = "#01a0e4"
magenta = "#a16a94"
cyan = "#b5e4f4"
white = "#a5a2a2"

[theme.bright]
black = "#5c5855"
red = "#db2d20"
green = "#01a252"
yellow = "#fded02"
blue = "#01a0e4"
magenta = "#a16a94"
cyan = "#b5e4f4"
white = "#f7f7f7"

[keys]
escape = "escape"
quit = "q"

[keys.view]
view_library = "1"
view_database = "2"
view_podcasts = "3"
open_config = "shift+C"
open_help = "control+h"

[keys.navigation]
up = "k"
down = "j"
left = "h"
right = "l"
goto_top = "g"
goto_bottom = "shift+G"

[keys.global_player]
toggle_pause = "space"
next_track = "n"
previous_track = "shift+N"
volume_up = "+"
volume_down = "-"
restart_track = "0"
seek_forward = "f"
seek_backward = "b"
speed_up = "control+f"
speed_down = "control+b"
toggle_prefetch = "control+g"
save_playlist = "control+s"

[keys.global_lyric]
adjust_offset_forwards = "shift+F"
adjust_offset_backwards = "shift+B"
cycle_frames = "shift+T"

[keys.library]
load_track = "l"
load_dir = "shift+L"
delete = "d"
yank = "y"
paste = "p"
cycle_root = "o"
add_root = "a"
remove_root = "shift+A"
search = "/"
youtube_search = "s"
open_tag_editor = "t"

[keys.playlist]
delete = "d"
delete_all = "shift+D"
shuffle = "r"
cycle_loop_mode = "m"
play_selected = "l"
search = "/"
swap_up = "shift+K"
swap_down = "shift+J"
add_random_songs = "s"
add_random_album = "shift+S"

[keys.database]
add_selected = "l"
add_all = "shift+L"

[keys.podcast]
search = "s"
mark_played = "m"
mark_all_played = "shift+M"
refresh_feed = "r"
refresh_all_feeds = "shift+R"
download_episode = "d"
delete_local_episode = "shift+D"
delete_feed = "x"
delete_all_feeds = "shift+X"

[keys.adjust_cover_art]
move_left = "control+shift+arrowleft"
move_right = "control+shift+arrowright"
move_up = "control+shift+arrowup"
move_down = "control+shift+arrowdown"
increase_size = "control+shift+pageup"
decrease_size = "control+shift+pagedown"
toggle_hide = "control+shift+end"

[keys.config]
save = "control+s"

[ytdlp]
extra_args = ""
```

Note that the only important bits in this config are the `[style.*]` fields, all others can be disregarded, but a full one has been posted just for completion.

</details>

theeasternfurry:
@hasezoey You're right, I guess the transparency effect can already be achieved from the start by setting background_color = "Reset" and an enum wasn't necessary. That said, I do find it a bit inconvenient having to set it 6 times across all the [style.*] sections. Would it make sense to have an easier way to do this like a single global option that applies it everywhere?

hasezoey:
> That said, I do find it a bit inconvenient having to set it 6 times across all the [style.*] sections.

That is true and will likely be the case for the near future, but i plan to change the whole theme / style thing to be a bit more general and easier to understand and use.

> Would it make sense to have an easier way to do this like a single global option that applies it everywhere?

It would likely make sense, but where would you put this and what kind of options would it have?

theeasternfurry:
@hasezoey For now, I can't think anything better solution but I guess this PR should be closed. Next time I will keep in mind before doing it. Maybe, the global option can be done after we have a new theme / style so it won't cause much trouble.

tramhao:
What I mean is, if the transparent is just setting Color:Reset as background, instead of setting this in code, we should set in theme files. Thanks.

PR #726, opened 2026-07-04
feat: enable macOS media key support via AppKit run loop

### LINKED ISSUE #640: Unable to use media control buttons on macOS
## Environment & Versions
- System: macos
- Termusic version: v0.12.1[c]
- Rust version: 1.89.0
- Cargo version: 1.89.0

## Description of the Problem
First I would like to thank you for such an awesome project.
I have installed it using the cargo and no errors. But media controls are not working for me. I tried closing all other players(Browser, YT Music, Apple music) but it still doesn't work. I also noticed that it doesn't show up in media control center in macOS top bar
My Macbook has Norwegian keyboard layout with English/Norsk switching in OS.

Image

RG-IL:
## Summary

Fixes macOS media keys (F-row play/pause/next/previous) not reaching termusic — instead they opened Apple Music or were ignored.

## Root cause

souvlaki registers `MPRemoteCommandCenter` handlers for media key events,
but macOS delivers those events exclusively through AppKit's main run
loop. Rust CLI binaries don't initialize an `NSApplication` run loop by
default, so the registered handlers never received callbacks.

Additionally, `MPRemoteCommandCenter` dispatches callbacks via GCD's main
dispatch queue, which executes blocks on the process's main thread. The
existing `#[tokio::main]` setup monopolized the main thread, preventing
GCD from dispatching the callbacks.

###  — macOS module
- `init_macos_main_thread()`: Initializes `NSApplication` with
  `ActivationPolicyAccessory` (no Dock icon, but can receive media
  events).
- `pump_run_loop()`: Pumps the main `CFRunLoop` in 50ms intervals,
  processing GCD blocks from `MPRemoteCommandCenter`.
- 5-second threshold on Previous: if >5s into the current track,
  restart it; if ≤5s, skip to the previous track (standard behaviour
  matching Spotify / Apple Music).

###  — Main thread restructure (macOS only)
- macOS: spawn the Tokio async runtime on a background thread, freeing
  the main thread to pump `CFRunLoop` for AppKit event dispatch.
- Other platforms: unchanged (`tokio::runtime::Runtime::new().block_on(actual_main())`
  instead of `#[tokio::main]`).

### 
- Added `objc = 0.2` gated behind `[target.'cfg(target_os = \"macos\")'.dependencies]`.

### Code style
- `#[allow(unsafe_code)]` on the macos module, matching the existing
  pattern used by the Windows module in the same file.
- `#[allow(unexpected_cfgs)]` at the playback crate root for the older
  `objc` crate's internal `cfg(cargo-clippy)` usage.

## Testing
1. Build with `cargo build --features mpv --release` on macOS.
2. Run the server, connect with `termusic --backend mpv`.
3. Press F7/F8/F9 (play/pause, next, previous) — should control termusic.
4. Quickly press Previous twice: first press restarts the current track
   (if >5s in), second press skips to the previous track.

Closes #...

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ Files:
 | :-------------------: | :-------------: |
 | `termusic-server.log` | The server logs |
 |  `termusic-tui.log`   |  The TUI logs   |
+| `termusic-cover`     |   MPRIS cover   |
 
 The default log level is `WARNING` (can be changed via [`RUST_LOG`](https://docs.rs/env_logger/latest/env_logger/#enabling-logging)).
 
--- a/playback/Cargo.toml
+++ b/playback/Cargo.toml
@@ -43,6 +43,9 @@ rodio.workspace = true
 tokio.workspace = true
 # soundtouch= { git = 'https://github.com/Drewol/soundtouch-rs.git' }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+objc2 = "0.6"
+
 [target.'cfg(target_os = "windows")'.dependencies]
 windows = { workspace = true, features = [
 	"Win32_Foundation",
--- a/playback/src/lib.rs
+++ b/playback/src/lib.rs
@@ -26,6 +26,8 @@ pub use backends::{Backend, BackendSelect};
 
 mod discord;
 mod mpris;
+#[cfg(target_os = "macos")]
+pub use mpris::macos;
 pub mod playlist;
 
 #[macro_use]
--- a/playback/src/mpris.rs
+++ b/playback/src/mpris.rs
@@ -395,3 +395,130 @@ mod windows {
     //     info!("Windows Event Queue Pump Ending");
     // }
 }
+
+#[cfg(target_os = "macos")]
+#[allow(unsafe_code)]
+pub mod macos {
+    //! macOS `AppKit` integration for media key support.
+    //!
+    //! `souvlaki`'s macOS backend registers `MPRemoteCommandCenter` handlers for
+    //! media key events, but macOS delivers those events through `AppKit`'s main run
+    //! loop. Rust CLI binaries don't initialize an `NSApplication` run loop by
+    //! default, so the registered handlers never receive events.
+    //!
+    //! Both [`init_macos_main_thread`] and [`pump_run_loop`] must be called from
+    //! the main thread. Apple's [`NSApplicationMain(_:_:)`] documentation
+    //! explicitly states: *"You must call this function from the main thread of
+    //! your application."* The Thread Safety Summary also notes that the main
+    //! thread is *"the one blocked in the `run` method of `NSApplication`"* and
+    //! that `NSRunLoop` is not thread-safe ([Thread Safety Summary]). In
+    //! practice, `MPRemoteCommandCenter` dispatches callbacks via GCD's main
+    //! queue which executes on the main thread, so event delivery breaks without
+    //! the main thread pumping the run loop.
+    //!
+    //! [`NSApplicationMain(_:_:)`]: https://developer.apple.com/documentation/appkit/nsapplicationmain(_:_:)/
+    //! [Thread Safety Summary]: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html#//apple_ref/doc/uid/10000057i-CH12-SW1
+    //!
+    //! This module provides:
+    //! - [`init_macos_main_thread`]: Initialize `NSApplication` (no Dock icon).
+    //! - [`pump_run_loop`]: Pump the main run loop so `AppKit` can dispatch events.
+    //! - [`run_with_run_loop`]: Convenience: spawn a closure on a background thread
+    //!   while pumping the run loop on the main thread.
+
+    use std::time::Duration;
+
+    use objc2::msg_send;
+    use objc2::runtime::{AnyClass, AnyObject};
+
+    /// `NSApplicationActivationPolicy.accessory` — run as a background process
+    /// with no Dock icon, but still able to receive media key events.
+    const NS_APPLICATION_ACTIVATION_POLICY_ACCESSORY: i64 = 1;
+
+    /// Initialize the macOS `AppKit` application on the main thread.
+    ///
+    /// Must be called from the main thread (the only thread that can run
+    /// `AppKit` per Apple's [Thread Safety Summary](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html))
+    /// before any `souvlaki`
+    /// `MediaControls` are attached. Sets the activation policy to
+    /// `.accessory`, meaning the app appears as a background process (no Dock
+    /// icon) but can still receive media key events via
+    /// `MPRemoteCommandCenter`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `NSApplication` class is not available at runtime (should
+    /// never happen on a real macOS system).
+    pub fn init_macos_main_thread() {
+        unsafe {
+            let cls = AnyClass::get(c"NSApplication").expect("NSApplication class not found");
+            let app: *mut AnyObject = msg_send![cls, sharedApplication];
+            let _: () =
+                msg_send![app, setActivationPolicy: NS_APPLICATION_ACTIVATION_POLICY_ACCESSORY];
+        }
+    }
+
+    /// Pump the main `CFRunLoop` until the sender is dropped or a message is
+    /// received (i.e. the background thread finished).
+    ///
+    /// Each iteration processes any pending events via `runUntilDate:` with
+    /// `distantPast` (non-blocking), then sleeps for 50 ms. This loop is
+    /// intended to run on the main thread while the Tokio runtime executes on a
+    /// background thread. When the background thread completes (or panics), the
+    /// sender is dropped and `recv_timeout` returns `Err`, causing this function
+    /// to return.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `NSRunLoop` or `NSDate` class is not available at runtime
+    /// (should never happen on a real macOS system).
+    pub fn pump_run_loop(done: &std::sync::mpsc::Receiver<()>) {
+        let rl_cls = AnyClass::get(c"NSRunLoop").expect("NSRunLoop class not found on macOS");
+        let date_cls = AnyClass::get(c"NSDate").expect("NSDate class not found on macOS");
+        loop {
+            match done.recv_timeout(Duration::from_millis(50)) {
+                Ok(()) | Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => break,
+                Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {}
+            }
+            unsafe {
+                let rl: *mut AnyObject = msg_send![rl_cls, mainRunLoop];
+                let distant_past: *mut AnyObject = msg_send![date_cls, distantPast];
+                let _: () = msg_send![rl, runUntilDate: distant_past];
+            }
+        }
+    }
+
+    /// Run a closure on a background thread while pumping the `AppKit` run loop
+    /// on the main thread.
+    ///
+    /// This is the simplest way to use the macOS media key support: call this
+    /// from `main()`, passing a closure that sets up and runs the Tokio runtime.
+    /// The closure runs on a background thread named "termusic-tokio", and the
+    /// main thread pumps `CFRunLoop` until the closure returns. `AppKit` and
+    /// `CFRunLoop` must run on the main thread per Apple's
+    /// [Thread Safety Summary](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the background thread cannot be spawned or panics.
+    pub fn run_with_run_loop<T>(f: impl FnOnce() -> T + Send + 'static) -> T
+    where
+        T: Send + 'static,
+    {
+        init_macos_main_thread();
+
+        let (tx, rx) = std::sync::mpsc::channel::<()>();
+
+        let handle = std::thread::Builder::new()
+            .name("termusic-tokio".into())
+            .spawn(move || {
+                let result = f();
+                let _ = tx.send(());
+                result
+            })
+            .expect("failed to spawn termusic-tokio thread");
+
+        pump_run_loop(&rx);
+        handle.join().expect("termusic-tokio thread panicked")
+    }
+}
+
--- a/playback/src/mpris.rs
+++ b/playback/src/mpris.rs
@@ -3,7 +3,6 @@ use std::{
     time::Duration,
 };
 
-use base64::Engine;
 use souvlaki::{MediaControlEvent, MediaControls, MediaMetadata, MediaPlayback, PlatformConfig};
 use termusiclib::{
     common::const_unknown::{UNKNOWN_ARTIST, UNKNOWN_TITLE},
@@ -58,18 +57,18 @@ impl Mpris {
     /// Set Mpris metadata based on the given track.
     pub fn set_track(&mut self, track: &Track) {
         let cover_art = match track.get_picture() {
-            Ok(v) => v.map(|v| {
-                format!(
-                    "data:{};base64,{}",
-                    v.mime_type().map_or_else(
-                        || {
-                            error!("Unknown mimetype for picture of track {track:#?}");
-                            "application/octet-stream"
-                        },
-                        |v| v.as_str()
-                    ),
-                    base64::engine::general_purpose::STANDARD_NO_PAD.encode(v.data())
-                )
+            Ok(v) => v.and_then(|v| {
+                let mut path = std::env::temp_dir();
+                path.push("termusic-cover");
+                if let Some(mime) = v.mime_type()
+                    && let Some(ext) = mime.ext()
+                {
+                    path.set_extension(ext);
+                }
+                std::fs::write(&path, v.data())
+                    .inspect_err(|e| error!("Saving cover to file failed: {e}"))
+                    .ok()?;
+                Some(format!("file://{}", path.display()))
             }),
             Err(err) => {
                 error!("Fetching the cover failed: {err:#?}");
@@ -374,26 +373,6 @@ mod windows {
     //         }
     //     }
     // }
-
-    // The following does not seem to be necessary, hence disabled
-    // /// Blockingly handle the windows event queue.
-    // ///
-    // /// This should be spawned on a extra thread
-    // pub fn pump_event_queue() {
-    //     use windows::Win32::UI::WindowsAndMessaging::{MSG, WM_QUIT, GetMessageW, TranslateMessage, DispatchMessageW};
-
-    //     let mut msg = MSG::default();
-
-    //     info!("Windows Event Queue Pump Starting");
-    //     unsafe {
-    //         while GetMessageW(&mut msg, None, 0, 0).as_bool() && msg.message != WM_QUIT {
-    //             debug!("Windows Message: {:#?}", msg);
-    //             let _ = TranslateMessage(&msg);
-    //             DispatchMessage(&msg);
-    //         }
-    //     }
-    //     info!("Windows Event Queue Pump Ending");
-    // }
 }
 
 #[cfg(target_os = "macos")]
@@ -521,4 +500,3 @@ pub mod macos {
         handle.join().expect("termusic-tokio thread panicked")
     }
 }
-
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -86,11 +86,13 @@ impl PlayerStats {
 }
 
 fn main() -> Result<()> {
+    #[cfg(target_os = "macos")]
+    let res = termusicplayback::macos::run_with_run_loop(actual_main);
+    #[cfg(not(target_os = "macos"))]
     let res = actual_main();
 
     trace!("Tokio Exited");
 
-    // print error to the log and then throw it
     if let Err(err) = res {
         error!("Error: {err:?}");
         return Err(err);
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+/graphify-out
 *.diff
 # ignore custom cargo configuration, like changing linker
 .cargo
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5383,6 +5383,7 @@ dependencies = [
  "gstreamer",
  "libmpv-sirno",
  "log",
+ "objc2",
  "parking_lot",
  "pathdiff",
  "pretty_assertions",


playback/src/lib.rs
```diff
@@ -1,3 +1,8 @@
+// The `objc` crate (v0.2, last updated 2020) uses `cfg(cargo-clippy)` internally
+// without declaring it, which triggers `unexpected_cfgs` on Rust >= 1.80.
+// All upstream objc calls are behind `#[cfg(target_os = "macos")]`.
+#![allow(unexpected_cfgs)]
```
hasezoey: Can this not be moved somewhere else instead of being crate-wide?

playback/src/mpris.rs
```diff
@@ -375,23 +380,75 @@ mod windows {
...
+    ///
+    /// Must be called from the main thread before any souvlaki `MediaControls`
+    /// are attached. Sets the activation policy to `.accessory`, meaning the
+    /// app appears as a background process (no Dock icon) but can still receive
+    /// media key events via `MPRemoteCommandCenter`.
+    pub fn init_macos_main_thread() {
+        unsafe {
+            let app: *mut objc::runtime::Object = msg_send![class!(NSApplication), sharedApplication];
+            // NSApplicationActivationPolicyAccessory = 1
+            let () = msg_send![app, setActivationPolicy: 1i64];
```
hasezoey: Is there no constant that could be used instead of a magic number (even if documented)?

server/src/server.rs
```diff
@@ -86,20 +86,45 @@ impl PlayerStats {
...
-        return Err(err);
-    }
+        if let Err(err) = res {
+            error!("Error: {err:?}");
+            return Err(err);
+        }
 
-    Ok(())
+        Ok(())
+    }
```
hasezoey, line 99: This duplicated some things that could be combined, even if a little more `#[cfg]` is necessary

hasezoey (changes requested):
Thanks for investigating what the issue is and trying to fix it.
As i cannot test it myself as i dont have a macos device, i would like some feedback from actual macos users.

Aside from that, please remove the changes that are beside the "fix macos media control" changes (ie your `Updated TUI to be like I wanted` commit).
Additionally, please remove the `5-second threshold on Previous: if >5s into the current track,` change as that is also beside the point and would make it inconsistent with the rest of termusic.
All of the changes removed can be another PR if you want.

playback/src/mpris.rs
```diff
@@ -375,23 +380,75 @@ mod windows {
...
+//     unsafe {
+//         while GetMessageW(&mut msg, None, 0, 0).as_bool() && msg.message != WM_QUIT {
+//             debug!("Windows Message: {:#?}", msg);
+//             let _ = TranslateMessage(&msg);
+//             DispatchMessage(&msg);
+//         }
+//     }
+//     info!("Windows Event Queue Pump Ending");
+// }
+}
```
hasezoey: Forgot to mention in the original review, please revert this change if you dont actually modify it.

RG-IL:
Sorry I'm new to this PR world so I didn't think that my commits after the PR was created will be in the PR. I will drop these commits. Thanks.

RG-IL:
I verified that it works well with the media buttons, but I haven't checked the Now Playing integration yet. I'll test that and report back

playback/src/mpris.rs
RG-IL: Will do it

playback/src/lib.rs
RG-IL: WIP

RG-IL: Hi, i checked and if it is not crate wide it does that:

```txt
warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:404:51
    |
404 |             let app: *mut objc::runtime::Object = msg_send![class!(NSApplication), sharedApplication];
    |                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `sel_impl` crate for guidance on how handle this unexpected cfg
    = help: the macro `sel_impl` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `#[warn(unexpected_cfgs)]` on by default
    = note: this warning originates in the macro `sel_impl` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:404:61
    |
404 |             let app: *mut objc::runtime::Object = msg_send![class!(NSApplication), sharedApplication];
    |                                                             ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `class` crate for guidance on how handle this unexpected cfg
    = help: the macro `class` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `class` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:405:22
    |
405 |             let () = msg_send![app, setActivationPolicy: NS_APPLICATION_ACTIVATION_POLICY_ACCESSORY];
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `sel_impl` crate for guidance on how handle this unexpected cfg
    = help: the macro `sel_impl` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `sel_impl` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:419:54
    |
419 |                 let rl: *mut objc::runtime::Object = msg_send![class!(NSRunLoop), mainRunLoop];
    |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `sel_impl` crate for guidance on how handle this unexpected cfg
    = help: the macro `sel_impl` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `sel_impl` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:419:64
    |
419 |                 let rl: *mut objc::runtime::Object = msg_send![class!(NSRunLoop), mainRunLoop];
    |                                                                ^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `class` crate for guidance on how handle this unexpected cfg
    = help: the macro `class` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `class` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:420:64
    |
420 |                 let distant_past: *mut objc::runtime::Object = msg_send![class!(NSDate), distantPast];
    |                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `sel_impl` crate for guidance on how handle this unexpected cfg
    = help: the macro `sel_impl` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `sel_impl` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:420:74
    |
420 |                 let distant_past: *mut objc::runtime::Object = msg_send![class!(NSDate), distantPast];
    |                                                                          ^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `class` crate for guidance on how handle this unexpected cfg
    = help: the macro `class` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `class` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition value: `cargo-clippy`
   --> playback/src/mpris.rs:421:26
    |
421 |                 let () = msg_send![rl, runUntilDate: distant_past];
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `all-backends`, `default`, `gst`, `mpv`, `rusty-libopus`, and `rusty-soundtouch`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `sel_impl` crate for guidance on how handle this unexpected cfg
    = help: the macro `sel_impl` may come from an old version of the `objc` crate, try updating your dependency with `cargo update -p objc`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: this warning originates in the macro `sel_impl` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Those  warnings are caused by upstream objc macro expansion in playback/src/mpris.rs. The class! and msg_send! macros expand through sel_impl, and that macro internally emits cfg(cargo-clippy), which triggers Rust’s unexpected_cfgs warnings at the expansion sites.

When these warnings are left unsuppressed, or only suppressed locally inside mpris.rs, the compiler still reports them because they originate from the external macro expansion context rather than from ordinary code in that function. I tried narrower allow attributes, but they were not sufficient, so the only practical fix in this crate is to add #![allow(unexpected_cfgs)] to playback/src/lib.rs.

This is a crate-local workaround for the termusic-playback crate only; it does not affect the rest of the workspace. It is necessary because the warnings come from an upstream macro implementation detail rather than from our own code, and a module-local allow does not reliably cover those macro expansions.

Once objc/sel_impl stops emitting cfg(cargo-clippy) internally, this crate-level workaround can be removed.

RG-IL:
I changed what you wanted and tested the now playing integration. It It works perfectly fine. You can skip forward, go to the previous song, and play and pause normally. Waiting to be merged...

playback/src/lib.rs
hasezoey: > Once objc/sel_impl stops emitting cfg(cargo-clippy) internally, this crate-level workaround can be removed.

`objc` is unlikely to update, looking at the "last updated over 6 years ago" and that this fix had already been merged but never released.

Maybe consider switching to `objc2` or `objr`?

server/src/server.rs
```diff
@@ -86,11 +86,41 @@ impl PlayerStats {
...
+    macos::init_macos_main_thread();
+
+    let done = Arc::new(AtomicBool::new(false));
+    let done_clone = done.clone();
+
+    let handle = std::thread::Builder::new()
+        .name("termusic-tokio".into())
+        .spawn(move || {
+            let rt = tokio::runtime::Runtime::new()?;
+            let result = rt.block_on(actual_main());
```
hasezoey: Couldnt the `#[tokio::main]` changes be reverted and instead use `let rt = actual_main()`?

playback/src/mpris.rs
```diff
@@ -374,24 +376,51 @@ mod windows {
...
-    //     unsafe {
-    //         while GetMessageW(&mut msg, None, 0, 0).as_bool() && msg.message != WM_QUIT {
-    //             debug!("Windows Message: {:#?}", msg);
-    //             let _ = TranslateMessage(&msg);
-    //             DispatchMessage(&msg);
-    //         }
-    //     }
-    //     info!("Windows Event Queue Pump Ending");
-    // }
+///
```
hasezoey:
```suggestion
```

playback/src/mpris.rs
```diff
@@ -374,24 +376,51 @@ mod windows {
...
+/// This module provides two functions:
+/// - [`init_macos_main_thread`]: Call once at startup to initialize `NSApplication`
+///   and set the activation policy to `.accessory` (no Dock icon).
+/// - [`pump_run_loop`]: Pump the main run loop so AppKit can dispatch event
+///   callbacks. Designed to be called from the main thread while a background
+///   thread runs the Tokio runtime.
+#[cfg(target_os = "macos")]
+#[allow(unsafe_code)]
+pub mod macos {
```
hasezoey, line 380: From what i can tell on a quick skim online, i dont see a explicit requirement that pumping the event loop requires to happen on the main thread, though likely requires the same thread as `NSAppliaction` is created (like in `init_macos_main_thread`?).

Though its not like i could test that as i dont have a macos system.

hasezoey (changes requested):
Thanks for resolving the changes, just some minor style nitpicks left (along with some extra testing).
Also please fix the lint (see failing CI).

Personally, i dont like things that require to run a specific thread, but if that is required to make this work, i have nothing against adding it (except are the things that are created on one thread and require to run on that thread (ex. a `MutexGuard`)).

server/src/server.rs
RG-IL: actual_main() returns a Future, not a Runtime, so let rt = actual_main()? wouldn't compile. The manual Runtime::new() + block_on() is needed here because the macOS path
  requires tokio to run on a non-main thread (the main thread must pump the AppKit run loop for media key events). The non-macOS path uses the same pattern for
  consistency.

hasezoey: > actual_main() returns a Future, not a Runtime

It is a future only if not annotated with `#[tokio::main]`. See [the documentation](https://docs.rs/tokio/latest/tokio/attr.main.html#set-the-name-of-the-runtime)

playback/src/mpris.rs
RG-IL: Both functions must run on the main thread — this is a hard AppKit requirement, not a design choice. Apple's NSApplication docs state that it's created via
  NSApplicationMain() which is called from main(), and NSRunLoop.mainRunLoop is by definition the main thread's run loop. Apple's Threading Programming Guide explicitly
  states: "The main thread of a Cocoa application also has a unique requirement in that it must run AppKit." https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html#//apple_ref/doc/uid/100000
        57i-CH12-SW1. MPRemoteCommandCenter dispatches media key
  callbacks through this run loop, so creating NSApplication on a non-main thread leads to events not being delivered.

hasezoey: > Both functions must run on the main thread — this is a hard AppKit requirement, not a design choice. Apple's NSApplication docs state that it's created via
> NSApplicationMain() which is called from main(), and NSRunLoop.mainRunLoop is by definition the main thread's run loop. Apple's Threading Programming Guide explicitly
> states: "The main thread of a Cocoa application also has a unique requirement in that it must run AppKit."

I cannot find this reference, the only thing i can find is the following text which does not indicate that it is strictly required on the main(entry) thread:
> Event Handling Restrictions
> The main thread is the one blocked in the run method of [NSApplication](https://developer.apple.com/documentation/appkit/nsapplication), usually invoked in an application’s main function. While the Application Kit continues to work if other threads are involved in the event path, operations can occur out of sequence

And i dont see `NSApplication` in the `Main Thread Only Classes` list (though i dont know if it inherits some of them).

---

I am not gonna argue further on this and trust you that what you say is correct as i cannot test this myself.

hasezoey (changes requested):
Some testing has now been done in #640 to confirm this PR is working.
So here is a recap of the things i would like still to be done before merging:
- consider switching to `objc2` or `objr` as `objc` is not updated anymore
- reduce the changes in `server.rs`

And in #640 it was noted that artwork display seemingly does not work, consider investigating this please.

RG-IL:
I've tested, and sometimes the artwork doesn't show up and sometimes it does. I will test it thoroughly tomorrow.

hasezoey:
> I've tested, and sometimes the artwork doesn't show up and sometimes it does

Do you have any indicator as to what tracks work and what dont?

But based on that *some* do work, it is not the issue that macos does not want it encoded as base64 inline-data instead of a file, right?

playback/src/mpris.rs
RG-IL: Thanks for double-checking the docs. I looked deeper and found the actual explicit references:
1. NSApplicationMain(_:_:) — Apple's documentation explicitly states: "You must call this function from the main thread of your application."
https://developer.apple.com/documentation/appkit/nsapplicationmain
2. Thread Safety Summary — Event Handling Restrictions: "The main thread of the application is responsible for handling events. The main thread is the one blocked in the run method of NSApplication, usually invoked in an application's main function."
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html
3. NSRunLoop is thread-unsafe (same doc, Thread-Unsafe Classes list) — mainRunLoop and runUntilDate: must be called from the thread that owns the run loop, which is the main thread.
So even though NSApplication isn't in the "Main Thread Only Classes" list (only NSCell/NSView are), NSApplicationMain() itself has an explicit main-thread requirement. And since NSApplication is created by NSApplicationMain() (called from main()), the shared application instance is bound to the main thread.
The updated doc comments now include the accurate references.

RG-IL:
> > I've tested, and sometimes the artwork doesn't show up and sometimes it does
> 
> Do you have any indicator as to what tracks work and what dont?
> 
> But based on that _some_ do work, it is not the issue that macos does not want it encoded as base64 inline-data instead of a file, right?

Root cause of intermittent artwork on macOS:
souvlaki v0.8.3's macOS backend (platform/macos/mod.rs) passes cover_url directly to NSImage initWithContentsOfURL:. This method does not officially support data: URIs — it expects file:// or https://. On some macOS versions or image sizes it may still succeed (Apple's internal URL loading may handle data URIs opportunistically), but the behavior is unreliable and undocumented.
termusic sends all artwork as data:image/{jpg,png};base64,... URIs. Our debug logging confirmed every track has found=true and set_metadata: success — the data is correctly extracted and passed to souvlaki. The failure is 100% in souvlaki's macOS backend not being designed for data URIs.
Fix options:
1. Upstream (souvlaki): Decode the base64 data URI in the macOS backend, create NSData from the bytes, then NSImage initWithData: instead of initWithContentsOfURL:
2. Workaround in termusic: Write cover art to a temp file and pass file:// URL; or bypass souvlaki for artwork on macOS entirely using objc2 directly (NSData → NSImage → MPMediaItemArtwork)

playback/src/lib.rs
RG-IL: done

RG-IL:
> 
>     2. Workaround in termusic: Write cover art to a temp file and pass file:// URL; or bypass souvlaki for artwork on macOS entirely using objc2 directly (NSData → NSImage → MPMediaItemArtwork)

fixed by writing the cover to tmp file, every time you change the song it replaces the image and the now plaing shows that images, work every time

hasezoey:
I am back a night later and see like 20 emails for this pr alone.
With the most recent change about 2 hours ago, so i am gonna ask: is this ready for review again?

RG-IL:
Sorry for the inconvenience, but i think so, thank you very much

playback/src/mpris.rs
```diff
@@ -90,6 +84,12 @@ impl Mpris {
         }
     }
 
+    fn cover_file_path() -> PathBuf {
+        let mut path = std::env::temp_dir();
+        path.push("termusic-cover.jpg");
+        path
+    }
```
hasezoey: Considering that this practically does not change across a single application run, this path could be cached.
Also docs are missing.
This should also be mentioned in the `Files` section of the README.

playback/src/mpris.rs
```diff
@@ -374,24 +374,123 @@ mod windows {
...
-    //     info!("Windows Event Queue Pump Ending");
-    // }
+/// macOS `AppKit` integration for media key support.
+///
+/// `souvlaki`'s macOS backend registers `MPRemoteCommandCenter` handlers for
+/// media key events, but macOS delivers those events through `AppKit`'s main run
+/// loop. Rust CLI binaries don't initialize an `NSApplication` run loop by
+/// default, so the registered handlers never receive events.
+///
+/// Both [`init_macos_main_thread`] and [`pump_run_loop`] must be called from
```
hasezoey: I am pretty sure those links are currently invalid.

playback/src/mpris.rs
```diff
@@ -374,24 +374,123 @@ mod windows {
...
+#[allow(unsafe_code)]
+pub mod macos {
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicBool, Ordering};
+    use std::time::Duration;
+
+    use objc2::msg_send;
+    use objc2::runtime::{AnyClass, AnyObject};
+
+    const NS_APPLICATION_ACTIVATION_POLICY_ACCESSORY: i64 = 1;
```
hasezoey, line 414: This should likely also have some documentation.

playback/src/mpris.rs
```diff
@@ -374,24 +374,123 @@ mod windows {
...
+
+    /// Initialize the macOS `AppKit` application on the main thread.
+    ///
+    /// Must be called from the main thread (the only thread that can run
+    /// `AppKit` per Apple's [Thread Safety Summary]) before any `souvlaki`
```
hasezoey: I am pretty sure this link is also invalid currently.

playback/src/mpris.rs
```diff
@@ -374,24 +374,123 @@ mod windows {
...
+    /// # Panics
+    ///
+    /// Panics if the `NSRunLoop` or `NSDate` class is not available at runtime
+    /// (should never happen on a real macOS system).
+    pub fn pump_run_loop(done: &AtomicBool) {
+        while !done.load(Ordering::SeqCst) {
+            unsafe {
+                let rl_cls = AnyClass::get(c"NSRunLoop").expect("NSRunLoop class not found");
+                let rl: *mut AnyObject = msg_send![rl_cls, mainRunLoop];
+                let date_cls = AnyClass::get(c"NSDate").expect("NSDate class not found");
```
hasezoey: Do the class-fetching need to happen every loop, or could they be done only once before the loop?

playback/src/mpris.rs
```diff
@@ -374,24 +374,123 @@ mod windows {
...
+    ///
+    /// Panics if the background thread cannot be spawned or panics.
+    pub fn run_with_run_loop<T>(f: impl FnOnce() -> T + Send + 'static) -> T
+    where
+        T: Send + 'static,
+    {
+        init_macos_main_thread();
+
+        let done = Arc::new(AtomicBool::new(false));
+        let done_clone = done.clone();
```
hasezoey: I dont think a AtomicBool is the correct choice here, as if tokio's thread panics (ie never gets to the `done.store(true)` part)), it may never actually exit or stop pumping the queue.

The workaround i would use it to setup a one-time channel and check for it to still be open every loop.
OR use tokio with a interval that pumps the queue, where the loop waits for a channel message.

server/src/server.rs
```diff
@@ -86,17 +86,22 @@ impl PlayerStats {
...
+        trace!("Tokio Exited");
 
-    Ok(())
+        if let Err(err) = res {
+            error!("Error: {err:?}");
+            return Err(err);
+        }
+
+        Ok(())
+    }
```
hasezoey, line 99:
```suggestion
        #[cfg(target_os = "macos")]
        let res = termusicplayback::macos::run_with_run_loop(actual_main);
        #[cfg(not(target_os = "macos"))]
        let res = actual_main();

        trace!("Tokio Exited");

        if let Err(err) = res {
            error!("Error: {err:?}");
            return Err(err);
        }

        Ok(())
```

This is what i meant all along, there is no need to diverge too much.

hasezoey (changes requested):
Looks better now that `objc2` is used, though there are still some style nitpicks, mainly documentation related.

hasezoey:
Btw, as a side-effect of using a file for the coverart in mpris again, is that artwork will likely work with GNOME again.

@Porkepix Could you try this PR on your linux machine and see if it still would cause the issue described in #731? This would at least rule-out / narrow down the issue GNOME is having.

RG-IL:
> Btw, as a side-effect of using a file for the coverart in mpris again, is that artwork will likely work with GNOME again.

As i do not have a Linux machine I cannot check. I will wait for a report from a GNOME user.

hasezoey:
> > Btw, as a side-effect of using a file for the coverart in mpris again, is that artwork will likely work with GNOME again.
> 
> As i do not have a Linux machine I cannot check. I will wait for a report from a GNOME user.

This was only meant as a side-note and not relevant to the PR being merged.

Porkepix:
> Btw, as a side-effect of using a file for the coverart in mpris again, is that artwork will likely work with GNOME again.
> 
> @Porkepix Could you try this PR on your linux machine and see if it still would cause the issue described in #731? This would at least rule-out / narrow down the issue GNOME is having.

Is this PR supposed to have any effect on the matter to specifically try this one?
Also note the crash reported in #731 is from `termusic` latest stable, 0.13.2, not master; so you could prefer to check this first?

I only build masters on a regular basis on my other Linux computer (better and more recent CPU, and 4x more RAM; on the one where I got the crash I tend to avoid due to how heavy it weighs regarding RAM usage).

hasezoey:
> I only build masters on a regular basis on my other Linux computer (better and more recent CPU, and 4x more RAM; on the one where I got the crash I tend to avoid due to how heavy it weighs regarding RAM usage).

Ah ok.

> Is this PR supposed to have any effect on the matter to specifically try this one?

Yes, because from what i remember on GNOME the artwork was also not working, which is the environment you reported to be using in #731 and because only some songs seemingly had the issue and some did not, so this would help narrow down (or rule-out) what parts cause GNOME to lock-up.

Porkepix:
So, building from this branch:
- `cargo build`, no issue.
- `cargo build --features cover` or `cargo build --features cover-ueberzug` > issue reproduced.

I can fairly claim issue comes from the code around ueberzug, and this with both `ueberzug` or `ueberzugpp` installed (I didn't tried if none is installed, didn't try to build without).

I didn't checked if it was fixing covers not showing for GNOME for other songs though, is it something needed too?

hasezoey:
> I can fairly claim issue comes from the code around ueberzug, and this with both ueberzug or ueberzugpp installed (I didn't tried if none is installed, didn't try to build without).

Thanks for narrowing the issue.

> I didn't checked if it was fixing covers not showing for GNOME for other songs though, is it something needed too?

Yes that would be great, but not necessary.

playback/src/mpris.rs
```diff
@@ -376,47 +384,50 @@ mod windows {
...
+    /// with no Dock icon, but still able to receive media key events.
     const NS_APPLICATION_ACTIVATION_POLICY_ACCESSORY: i64 = 1;
 
     /// Initialize the macOS `AppKit` application on the main thread.
     ///
     /// Must be called from the main thread (the only thread that can run
-    /// `AppKit` per Apple's [Thread Safety Summary]) before any `souvlaki`
+    /// `AppKit` per Apple's
+    /// <https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html>)
+    /// before any `souvlaki`
```
hasezoey: For some reason github does not give me the suggestion option.

This link should likely not be a plain link. (use `[]()` instead of `<>`)

playback/src/mpris.rs
```diff
@@ -467,7 +480,8 @@ pub mod macos {
     /// from `main()`, passing a closure that sets up and runs the Tokio runtime.
     /// The closure runs on a background thread named "termusic-tokio", and the
     /// main thread pumps `CFRunLoop` until the closure returns. `AppKit` and
-    /// `CFRunLoop` must run on the main thread per Apple's [Thread Safety Summary].
+    /// `CFRunLoop` must run on the main thread per Apple's
+    /// <https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html>.
```
hasezoey: The same here, `[]()` should be used over `<>`

playback/src/mpris.rs
```diff
@@ -435,28 +446,30 @@ pub mod macos {
...
     ///
     /// Panics if the `NSRunLoop` or `NSDate` class is not available at runtime
     /// (should never happen on a real macOS system).
-    pub fn pump_run_loop(done: &AtomicBool) {
-        while !done.load(Ordering::SeqCst) {
+    pub fn pump_run_loop(done: &std::sync::mpsc::Receiver<()>) {
+        let rl_cls =
+            AnyClass::get(c"NSRunLoop").expect("NSRunLoop class not found on macOS");
+        let date_cls = AnyClass::get(c"NSDate").expect("NSDate class not found on macOS");
+        while done.recv_timeout(Duration::from_millis(50)).is_err() {
```
hasezoey: As written, this does not actually solve the panic issue.

Either explicitly match on the error type being `Timeout` instead of `Disconnected`, or use a tokio [oneshot](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html) channel, which provides a `Receiver::is_terminated()` check. (std will also have that, but there it is nightly-only)
(or go with a normal mpsc channel and have tokio be the interval)

playback/src/mpris.rs
```diff
@@ -19,6 +19,16 @@ pub struct Mpris {
     pub rx: Receiver<MediaControlEvent>,
 }
 
+/// Path to the temporary cover art file exposed via MPRIS.
+///
+/// The path is computed once and cached, since `temp_dir()` does not change
+/// during a process lifetime.
+static COVER_FILE_PATH: std::sync::LazyLock<PathBuf> = std::sync::LazyLock::new(|| {
+    let mut path = std::env::temp_dir();
+    path.push("termusic-cover.jpg");
```
hasezoey: Are we always converting the image to be JPG? If not, this extension is wrong.
Consider to remove the extension and have a plain file that is updated, or dynamically set the extension.

hasezoey (changes requested):

playback/src/mpris.rs
```diff
@@ -463,7 +453,11 @@ pub mod macos {
     pub fn pump_run_loop(done: &std::sync::mpsc::Receiver<()>) {
         let rl_cls = AnyClass::get(c"NSRunLoop").expect("NSRunLoop class not found on macOS");
         let date_cls = AnyClass::get(c"NSDate").expect("NSDate class not found on macOS");
-        while done.recv_timeout(Duration::from_millis(50)).is_err() {
+        loop {
+            match done.recv_timeout(Duration::from_millis(50)) {
+                Ok(()) | Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => break,
+                Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {}
```
hasezoey, line 459: Import `std::sync::mpsc::RecvTimeoutError`

playback/src/mpris.rs
```diff
@@ -69,8 +58,14 @@ impl Mpris {
...
-                let path = Self::cover_file_path();
-                std::fs::write(path, v.data())
+                let mut path = std::env::temp_dir();
+                path.push("termusic-cover");
+                if let Some(mime) = v.mime_type()
+                    && let Some(ext) = mime.ext()
+                {
+                    path.set_extension(ext);
+                }
+                std::fs::write(&path, v.data())
```
hasezoey, line 68: I just remembered we are also doing a similar thing for the ueberzug case already and there always set the JPG extension.

This is just a side-note and does not require any change in this PR.

server/src/server.rs
hasezoey: Still a nitpick

hasezoey (changes requested):
Looks almost ready, still some minor style nitpicks.

Once done, please reduce the commits down to 2 (one for the macos change and one for the coverart change)
Note that the commit message style is meant to include the *crate* that is modified in the scope field. (see [CONTRIBUING](https://github.com/tramhao/termusic/blob/master/CONTRIBUTING.md#commit-messages), also see recent past commits for reference)

playback/src/mpris.rs
hasezoey: Upon further investigation (which i overlooked), we do always set the JPG extension, but also always convert the image to JPG.

playback/src/mpris.rs
RG-IL: Fixed — imported RecvTimeoutError and used short form in the match arms.

server/src/server.rs
RG-IL: Applied your suggestion — simplified main() to share error handling between both cfg paths.

playback/src/mpris.rs
hasezoey: This does not look applied

playback/Cargo.toml
```diff
@@ -43,6 +43,9 @@ rodio.workspace = true
 tokio.workspace = true
 # soundtouch= { git = 'https://github.com/Drewol/soundtouch-rs.git' }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+objc2 = "0.6"
```
hasezoey, line 47: Any dependency version should be defined (at least currently) in the workspace's Cargo.toml

README.md
```diff
@@ -187,6 +187,7 @@ Files:
 | :-------------------: | :-------------: |
 | `termusic-server.log` | The server logs |
 |  `termusic-tui.log`   |  The TUI logs   |
+| `termusic-cover`     |   MPRIS cover   |
```
hasezoey, line 190:
```suggestion
| `termusic-cover.EXT`   |   MPRIS cover   |
```

hasezoey (approved):
Looks great, minor style nitpicks (which can be fixed after this PR), though i will re-test tomorrow and then merge.

RG-IL:
so should i make the changes or no?

PR #730, opened 2026-07-06
implement dir entries filter by the audio mime type #727

### LINKED ISSUE #727: Hide directory entries without playback
I have some files in music collection such as .torrent and .!qB (incompleted downloads)

Image

What about hiding them? I think [mime_guess](https://crates.io/crates/mime_guess) crate would help with its MUSIC const as the filter for listing.

oooo-ps:
fixes #727

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2921,6 +2921,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
 
 [[package]]
+name = "mime_guess"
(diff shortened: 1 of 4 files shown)


hasezoey (commented):
Thanks for the PR, the code looks OK, though i dont like that this is not configurable (including the already existing hidden files being filtered out).
Though i know adding customization is more work.

@tramhao what do you think about this?

oooo-ps:
Please complete the config option if it wanted, because I'm beginner in this app and my computer is super slow for active development.

One more idea - is to hide empty dirs and dirs that does not include audio files to play (for example, incomplete downloads like `.!qB` in my case) anyway it's another story as I'm already happy with the current filter impl.

Issue #671, opened 2026-03-06
Improve navigation in the color pane for configuration


Porkepix:
## Description of the Problem

Build on current master, in the Themes and Colors tab in the config editor, a couple of things makes it a little bit rough or counter-intuitive. I don't necessarily have ideas for them, but think it's worth mentioning:

- Tab to move between tabs also move between the left and right side of this tab, that's unsettling;
- In the right part user can only move between items using up and down arrow and only on some of the items, their vim equivalents (`j` and `k`) although those doesn't work on items with an input field where inputs are taken immediately without having the user press `enter`.
  - We should be able to use vim-like movement on all of them or none;
  - Ideally it'd be best to be able to move on left and right items with the corresponding movements. But I know there's a difficulty here as the layout is dynamic depending on the terminal size, and items are not always on the same line or column;
- In the right part, we move using up and down but instead of cycling, if we press `up` from the first item or `down` from the last item, we're sent back to the column on the left, which is quite unexpected.

grishmadev:
we can try adding an offset to the whenever the user presses up or down,
in this file: https://github.com/tramhao/termusic/blob/master/tui/src/ui/components/config_editor/general.rs

we can do something like:
```rs
impl AppComponent<Msg, UserEvent> for MusicDir {
    fn on(&mut self, ev: &Event<UserEvent>) -> Option<Msg> {
        handle_input_ev(
            &mut self.component,
            ev,
            &self.config.read().settings.keys,
            // skipping the whole line and landing directly below the focused block, example - 7 blocks of key config can fit in a row
            Msg::ConfigEditor(ConfigEditorMsg::General(KFMsg::Skip(7))),
            Msg::ConfigEditor(ConfigEditorMsg::General(KFMsg::Skip(-7))),
            // for right key
            Msg::ConfigEditor(ConfigEditorMsg::General(KFMsg::Next)),
            // for left key
            Msg::ConfigEditor(ConfigEditorMsg::General(KFMsg::Previous)),
        )
    }
}
// this isnt accurate and complications can be handled carefully later
```
for this to work, we need to know the dimensions of the screen or how many blocks can fit, which i think can be accessed by tuirealm library, i may be wrong.

hasezoey:
I dont know if i understand your comment correctly, do you mean to add a way to figure out how many blocks in the grid are in a given line and for up/down movement to skip that amount?

[...]

PR #734, opened 2026-07-08
fix: cycle up/down focus within color pane items instead of wrapping into the theme select table

mvanhorn:
## Summary
In the config editor's "Themes and Colors" tab, Up/Down now cycles within the right-hand color-item grid instead of throwing focus back to the theme-select table when you hit the first or last item. Tab/BackTab traversal of the full focus order is unchanged, so the deliberate table-to-grid entry points still work.

## Why this matters
The reporter's most concrete complaint on #671 was exactly this: Up from the first color item or Down from the last lands on the theme table in the other pane. Collaborator hasezoey confirmed the behavior is "not intuitive and inconsistent". The cause is that `THEME_FOCUS_ORDER` is one flat array whose element 0 is `ThemeSelectTable`, and `set_next_in_focus_array` wraps Next/Previous around the whole array, so arrow-driven moves cross the pane boundary. This scopes to the confirmed complaint and does not redesign Tab semantics or grid row-aware skipping, which the thread has not converged on.

## Changes
- New `ConfigEditorMsg::ThemeColorItem(KFMsg)` variant routes arrow-driven moves to a handler that cycles over the color-item sub-slice `&THEME_FOCUS_ORDER[1..]`, never landing on `ThemeSelectTable`.
- Only the Up/Down emitters in `CEColorSelect::on` (closed-select state) and `ConfigInputHighlight::on` switch to the new variant; Tab/BackTab arms keep emitting `Theme(KFMsg)`.

## Testing
Inline `#[cfg(test)]` tests in `update.rs` cover sub-slice cycling (Next from last color item wraps to first, Previous from first wraps to last, table never entered). `cargo clippy` and `cargo build` pass.

re #671

### DIFF
--- a/tui/src/ui/components/config_editor/update.rs
+++ b/tui/src/ui/components/config_editor/update.rs
@@ -116,6 +116,7 @@ impl Model {
             // Focus handling
             ConfigEditorMsg::General(msg) => self.update_general(msg),
             ConfigEditorMsg::Theme(msg) => self.update_theme(msg),
+            ConfigEditorMsg::ThemeColorItem(msg) => self.update_theme_color_item(msg),
             ConfigEditorMsg::KeyFocusGlobal(msg) => self.update_key_focus_global(msg),
             ConfigEditorMsg::KeyFocusOther(msg) => self.update_key_focus_other(msg),
             ConfigEditorMsg::ChangeLayout(msg) => self.change_layout(msg),
@@ -210,6 +211,17 @@ impl Model {
     /// Handle focus of the "Theme" tab
     fn update_theme(&mut self, msg: KFMsg) {
         set_next_in_focus_array(self, msg, THEME_FOCUS_ORDER, |id| {
+            if let IdConfigEditor::Theme(id) = id {
+                Some(id)
+            } else {
+                None
+            }
+        });
+    }
+
+    /// Handle focus inside the "Theme" tab's color-item pane
+    fn update_theme_color_item(&mut self, msg: KFMsg) {
+        set_next_in_focus_array(self, msg, &THEME_FOCUS_ORDER[1..], |id| {
             if let IdConfigEditor::Theme(id) = id {
                 Some(id)
             } else {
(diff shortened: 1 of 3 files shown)


Porkepix:
It seems there are other places affected by the same navigation issue such as at least the database view; maybe it could also handle this case?

tui/src/ui/components/config_editor/update.rs
```diff
@@ -511,29 +487,126 @@ where
...
 
+    let focus = next_in_focus_array(msg, array, focus_elem)?;
+
+    let id = focus.into();
+    let _ = model.app.active(&Id::ConfigEditor(id));
+
+    Some(id)
+}
+
+fn next_in_focus_array<T>(msg: KFMsg, array: &[T], focus_elem: Option<T>) -> Option<T>
```
hasezoey, line 508: Docs are missing.

hasezoey (changes requested):
Looks good, though needs some new documentation in `THEME_FOCUS_ORDER` that some entries may be skipped. Alternatively, add a new constant that sets a index into `THEME_FOCUS_ORDER` right below it with documentation to have it in a single place. (or just have 2 arrays altogether).

Note that i changed the `fixes` to `re` as this does not entirely address what was discussed in #671.

mvanhorn:
Added the note on `THEME_FOCUS_ORDER` documenting that focus handlers may cycle within a subslice so entries can be skipped, with the color-item pane's `ThemeSelectTable` as the example. Kept scope to this pane - the database view case @Porkepix mentioned looks real but is a separate handler, happy to do it as a follow-up PR.

mvanhorn:
Lint addressed in 8bac4e6. Verified locally: cargo fmt --check clean, cargo clippy on the tui crate clean, cargo test passing.

hasezoey (approved):
Looks good to me now.

I will test and merge this later when i have a little more time.

The Lint failure from clippy regarding `manual filter impl` may be ignored as this is fixed in #737.

mvanhorn:
Appreciate the merge, @hasezoey - color pane focus cycles within items as expected now.

mvanhorn:
Thanks @hasezoey! Focus cycling within the color pane beats wrapping into the wrong pane.

PR #738, opened 2026-07-11
fix(playback::mpris): use correct return type for setActivationPolicy msg_send

RG-IL:
Fixes a runtime panic on macOS where `setActivationPolicy:` was declared with return type `()` (void) but the Objective-C method returns `BOOL`. The Objective-C runtime detects the type mismatch and panics on startup.

The fix changes the return type annotation from `()` to `bool` to match the actual Objective-C method signature.

### DIFF
--- a/playback/src/mpris.rs
+++ b/playback/src/mpris.rs
@@ -431,7 +431,7 @@ pub mod macos {
         unsafe {
             let cls = AnyClass::get(c"NSApplication").expect("NSApplication class not found");
             let app: *mut AnyObject = msg_send![cls, sharedApplication];
-            let _: () =
+            let _: bool =
                 msg_send![app, setActivationPolicy: NS_APPLICATION_ACTIVATION_POLICY_ACCESSORY];
         }
     }


RG-IL:
@hasezoey Sorry, I haven't caught it in #726  As I used a custom version for myself and didn't properly test the final version of #726. Sorry for the inconvenience.

playback/src/mpris.rs
```diff
@@ -431,7 +431,7 @@ pub mod macos {
         unsafe {
             let cls = AnyClass::get(c"NSApplication").expect("NSApplication class not found");
             let app: *mut AnyObject = msg_send![cls, sharedApplication];
-            let _: () =
+            let _: bool =
```
hasezoey, line 434:
```suggestion
            let _ =
```

How about we just dont declare any type if we dont care about the return value, or is it necessary in this case?

RG-IL: No, let _ won't compile — msg_send! in the objc crate needs a concrete return type to generate the correct message-send calling convention. The type annotation is required.

hasezoey (approved):
If that is required, then sure.

For reference [NSApplication/setActivationPolicy(_:)](https://developer.apple.com/documentation/appkit/nsapplication/setactivationpolicy(_:)).

Issue #699, opened 2026-04-14
add key shortcut to restart track


henri:
## Describe what you want

key binding to rewind the current track to the start playback from the beginning.

## Do you have already an idea for the implementation?

No... maybe there is a way of doing this when restart position is enabled. Open to suggestions.

Thanks for this great player :)

hasezoey:
Any suggestions on what the default keybind should be?

For now, keybindings cannot be empty in termusic.

henri:
perhaps '0', control-leftarrow or even left arrow if the track is selected? Open to better ideas.

grishmadev:
I would like to work on this issue. is it still available?

hasezoey:
> is it still available?

As long as it is open and does not have a PR associated with it (which it currently does not).

grishmadev:
@hasezoey hey, i sent you an email, can you go through it and help me. its regarding this update

Porkepix:
To note, in many players the "previous track" button/shortcut is used for this if the play time is at more than just a couple of seconds.
I'm not sure it's the best idea, but that's clearly a common way to handle it.

hasezoey:
I dont know why this is still open when #717 had been merged, so i will close this for now.

---

> To note, in many players the "previous track" button/shortcut is used for this if the play time is at more than just a couple of seconds.

I am aware of that, and that would likely be good to at least have as a option, but now a dedicated way to restart a track [regardless of its played time] has been added.

henri:
Thank you for your work on this featuere. Nice press 0 to restart rack :100: makes sense!

PR #739, opened 2026-07-11
feat(playback): restart current track on SkipPrevious when past configurable threshold

RG-IL:
Adds a new config option previous_track_threshold (0-10s, default 5,
0=disabled). When the current track position exceeds the threshold,
pressing Previous restarts the current track instead of going to the
previous track.

The logic lives in player.previous() so it works seamlessly with all
inputs — macOS media keys, TUI keybindings, and any other caller.

Adds PreviousTrackThreshold newtype with serde support, v1 config
conversion, and TUI config editor option in the General tab.

re #699

### DIFF
--- a/lib/src/config/v2/server/mod.rs
+++ b/lib/src/config/v2/server/mod.rs
@@ -77,6 +77,61 @@ pub enum ScanDepth {
 /// What determines a long track length in seconds, 10 minutes
 const LONG_TRACK_TIME: u64 = 600; // 60 * 10
 
+/// How many seconds into a track before Command "Previous Track" restarts the current track vs going to the previous track.
+/// Range: 0-10, where 0 disables the feature (always goes to previous track).
+/// Default: 5 seconds.
+///
+/// If the current track position is past this threshold, pressing "Previous Track" will restart the current track.
+/// If before the threshold, it goes to the previous track.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(try_from = "u8", into = "u8")]
+pub struct PreviousTrackThreshold(u8);
+
+impl PreviousTrackThreshold {
+    /// Create a new `PreviousTrackThreshold`, clamping the value to 0-10.
+    #[must_use]
+    pub fn new(v: u8) -> Self {
+        Self(v.clamp(0, 10))
+    }
+
+    #[must_use]
+    pub fn get(&self) -> u64 {
+        u64::from(self.0)
+    }
+
+    /// Returns `true` if the threshold feature is enabled (value > 0).
+    #[must_use]
+    pub fn is_enabled(&self) -> bool {
+        self.0 > 0
+    }
+}
+
+impl Default for PreviousTrackThreshold {
+    fn default() -> Self {
+        Self(0)
+    }
+}
+
+impl TryFrom<u8> for PreviousTrackThreshold {
+    type Error = String;
+
+    fn try_from(v: u8) -> Result<Self, Self::Error> {
+        if v > 10 {
+            Err(format!(
+                "previous_track_threshold must be between 0 and 10, got {v}"
+            ))
+        } else {
+            Ok(Self(v))
+        }
+    }
+}
+
+impl From<PreviousTrackThreshold> for u8 {
+    fn from(v: PreviousTrackThreshold) -> Self {
+        v.0
+    }
+}
+
 /// Seek amount maybe depending on track length
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 #[serde(untagged)]
@@ -283,6 +338,11 @@ pub struct PlayerSettings {
     pub gapless: bool,
     /// How much to seek on a seek event
     pub seek_step: SeekStep,
+    /// Threshold for Previous Track behavior, in seconds (0-10).
+    ///
+    /// If the current track position is past this threshold, pressing "Previous Track" restarts the current track.
+    /// If before the threshold, it goes to the previous track. Set to 0 to disable (always goes to previous track).
+    pub previous_track_threshold: PreviousTrackThreshold,
 
     /// Controls if support via Media-Controls (like mpris on linux) is enabled
     pub use_mediacontrols: bool,
@@ -321,6 +381,7 @@ impl Default for PlayerSettings {
             speed: 10,
             gapless: true,
             seek_step: SeekStep::default(),
+            previous_track_threshold: PreviousTrackThreshold::default(),
 
             use_mediacontrols: true,
             set_discord_status: true,
@@ -519,8 +580,8 @@ mod v1_interop {
 
     use super::{
         Backend, ComSettings, LoopMode, NonZeroU8, NonZeroU32, PlayerSettings, PodcastSettings,
-        PositionYesNo, PositionYesNoLower, RememberLastPosition, ScanDepth, SeekStep,
-        ServerSettings, backends::BackendSettings,
+        PositionYesNo, PositionYesNoLower, PreviousTrackThreshold, RememberLastPosition, ScanDepth,
+        SeekStep, ServerSettings, backends::BackendSettings,
     };
     use crate::config::{
         v1,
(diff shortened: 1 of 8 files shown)


lib/src/config/v2/server/mod.rs
```diff
@@ -77,6 +77,58 @@ pub enum ScanDepth {
...
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let v: u8 = serde::Deserialize::deserialize(deserializer)?;
+        if v > 10 {
+            return Err(serde::de::Error::custom(
+                "previous_track_threshold must be between 0 and 10",
+            ));
+        }
+        Ok(Self(v))
+    }
+}
```
hasezoey: Consider adding a simple note why a manual ser/de impl over `#[serde(transparent)]` is necessary (from my quick glance it is for validation)
Maybe consider using a `#[serde(try_from = u8)]` and add a simple `impl TryFrom<u8> for TYPE`.

lib/src/config/v2/server/mod.rs
```diff
@@ -283,6 +335,11 @@ pub struct PlayerSettings {
     pub gapless: bool,
     /// How much to seek on a seek event
     pub seek_step: SeekStep,
+    /// Threshold for Previous Track behavior, in seconds (1-10).
```
hasezoey:
```suggestion
    /// Threshold for Previous Track behavior, in seconds (0-10).
```

Also a note again that `0` disables this would be great here.

lib/src/config/v2/server/mod.rs
```diff
@@ -283,6 +335,11 @@ pub struct PlayerSettings {
...
+    ///
+    /// If the current track position is past this threshold, Previous restarts the current track.
+    /// If before the threshold, it goes to the previous track.
+    pub previous_track_threshold: PreviousTrackThreshold,
```
hasezoey:
```suggestion
    /// If the current track position is past this threshold, Command "Previous Track" restarts the current track.
    /// If before the threshold, it goes to the previous track.
    pub previous_track_threshold: PreviousTrackThreshold,
```

Maybe a better name would be `restart_track_threshold`?

lib/src/config/v2/server/mod.rs
```diff
@@ -77,6 +77,58 @@ pub enum ScanDepth {
 /// What determines a long track length in seconds, 10 minutes
 const LONG_TRACK_TIME: u64 = 600; // 60 * 10
 
+/// How many seconds into a track before Previous Track restarts the current track vs going to the previous track.
```
hasezoey:
```suggestion
/// How many seconds into a track before Command "Previous Track" restarts the current track vs going to the previous track.
```

playback/src/lib.rs
```diff
@@ -587,8 +587,21 @@ impl GeneralPlayer {
         }
     }
 
-    /// Switch & Play the previous track in the playlist
+    /// Switch & Play the previous track in the playlist.
+    ///
+    /// If the current track position is past the configured `previous_track_threshold`,
+    /// the current track is restarted instead of switching to the previous track.
+    /// Set `previous_track_threshold` to 0 in the config to disable this behavior
+    /// (always goes to the previous track).
```
hasezoey:
```suggestion
    /// Set `previous_track_threshold` to 0 in the config to disable this behavior,
    /// which will then always go to the previous track.
```

tui/src/ui/components/config_editor/view.rs
```diff
@@ -790,6 +793,18 @@ impl Model {
...
 
+        if let Ok(State::Single(StateValue::String(threshold_str))) = self.app.state(
+            &Id::ConfigEditor(IdConfigEditor::General(IdCEGeneral::PreviousTrackThreshold)),
+        ) && let Ok(threshold) = threshold_str.parse::<u8>()
+        {
+            if threshold <= 10 {
+                config_server.settings.player.previous_track_threshold =
+                    PreviousTrackThreshold::new(threshold);
+            } else {
+                bail!(" Previous track threshold must be between 0 and 10.");
```
hasezoey:
```suggestion
                bail!(" Previous track threshold must be between 0 and 10. ");
```

tui/src/ui/components/config_editor/general.rs
```diff
@@ -1019,6 +1019,56 @@ impl AppComponent<Msg, UserEvent> for PlayerBackend {
...
+}
+
+impl ConfigPreviousTrackThreshold {
+    pub fn new(config: CombinedSettings) -> Self {
+        let component = {
+            let config_tui = config.tui.read();
+            common_input_comp(&config_tui, " Previous Track Threshold (0-10s, 0=disabled): ")
+                .input_type(InputType::UnsignedInteger)
+                .placeholder(LineStatic::styled(
+                    "5",
```
hasezoey: I would recommend instead of a number to repeat the `0-10s, 0=disabled` text, WDYT?

tui/src/ui/components/config_editor/general.rs
```diff
@@ -1019,6 +1019,56 @@ impl AppComponent<Msg, UserEvent> for PlayerBackend {
...
+    component: Input,
+    config: SharedTuiSettings,
+}
+
+impl ConfigPreviousTrackThreshold {
+    pub fn new(config: CombinedSettings) -> Self {
+        let component = {
+            let config_tui = config.tui.read();
+            common_input_comp(&config_tui, " Previous Track Threshold (0-10s, 0=disabled): ")
+                .input_type(InputType::UnsignedInteger)
```
hasezoey: This should also set a [`input_len`](https://docs.rs/tui-realm-stdlib/latest/tui_realm_stdlib/components/struct.Input.html#method.input_len). (i dont know if this also applies in `InputType::Custom`)
Also should likely make use of [`InputType::Custom`](https://docs.rs/tuirealm/4.1.0/tuirealm/props/enum.InputType.html#variant.Custom) to validate against the `0-10` limit, instead of just doing it on save.

playback/src/lib.rs
```diff
@@ -587,8 +587,21 @@ impl GeneralPlayer {
...
     pub fn previous(&mut self) {
+        self.player_save_last_position();
```
hasezoey: This should be moved after the restart track code.

hasezoey (changes requested):
Looks great, aside from some minor styling / documentation nitpicks.

tui/src/ui/components/config_editor/view.rs
```diff
@@ -790,6 +793,18 @@ impl Model {
...
+        if let Ok(State::Single(StateValue::String(threshold_str))) = self.app.state(
+            &Id::ConfigEditor(IdConfigEditor::General(IdCEGeneral::PreviousTrackThreshold)),
+        ) && let Ok(threshold) = threshold_str.parse::<u8>()
+        {
+            if threshold <= 10 {
+                config_server.settings.player.previous_track_threshold =
+                    PreviousTrackThreshold::new(threshold);
+            } else {
+                bail!(" Previous track threshold must be between 0 and 10. ");
+            }
```
hasezoey: How about we use the the `try_from(u8)` here too instead of checking the same condition? (though feel free to modify the error message in this case)

tui/src/ui/components/config_editor/general.rs
```diff
@@ -1019,6 +1019,75 @@ impl AppComponent<Msg, UserEvent> for PlayerBackend {
...
 
+fn threshold_valid(s: &str) -> bool {
+    if let Ok(v) = s.parse::<u8>() {
+        v <= 10
+    } else {
+        false
+    }
+}
+
+fn threshold_char_valid(s: &str, c: char) -> bool {
```
hasezoey, line 1038: Some documentation here would be good.

hasezoey (commented):
Some minor style nitpicks, otherwise looks ready.

For the future, might i recommend to use a separate branch than the fork's master?

RG-IL:
hi, I just wanted to say thanks for all of the support and the maintenance of this project. This is an awesome project that I use pretty much every day, and I loved working with you. Thank you very much.

hasezoey (approved):

hasezoey:
I just noticed that in the duration of this PR, you have changed the default to be `0s`, was that intentional?

RG-IL:
> I just noticed that in the duration of this PR, you have changed the default to be `0s`, was that intentional?

Yes as at the beginning I did 5 seconds because that's what I wanted but then realized that right now the default is 0. I should probably change it to what the default was already to not surprise users that update and don't understand why when they use previous song it's just starting from the beginning of the current one

PR #737, opened 2026-07-09
Some Styling cleanup

hasezoey:
This PR focuses on some styling issues:
- fix some styling issues from #726 
- sort workspace dependency list to be better understandable at a glance
- remove dependencies from `lib` which are not actually used
- remove a intermediate `lofty::Picture` parse, which could be directly parsed into `image`
- fix rust 1.97 styling warnings / errors

PR #740, opened 2026-07-12
Reduce noise & duplication for drawing in config editor "General"

hasezoey:
This PR changes the drawing order for config editor tab "General" to be based on the focus array, as for this case anything that is not in there should also not be drawn. Also reduces changing noise when adding / removing extra fields, as can be seen in #739.
Additionally, applies some style fixes that were missed in #739. To prevent this in the future, this also contains a workflow change to always run both "clippy" and "fmt" steps, even if one fails to report the status of both.


hasezoey:
To fix all issues at once, this PR has been combined into #737

PR #743, opened 2026-07-14
feat: playlist sort — DB-dependent criteria (most-played, recency, frecency)

### LINKED ISSUE #709: making the TUI have sorting by date added/ date modified
im not sure if that is already possible

a keabind to sort playlist files in date modified (old to new / new to old) with a keabind maybe maybe tyty

RG-IL:
This is **part 2 of 2** of the playlist sort feature. The base criteria (alphabetical, date-added, duration) and the `SortPlaylist` RPC it builds on live in PR #746.

Adds the **DB-dependent** sort criteria, which need new play statistics:

- **Most-played** (`f` / `F`) — most / least started
- **Recency** (`c` / `C`) — most / least recently started
- **Frecency** (`u` / `U`) — zoxide-style score weighting recency × play count (`×4` <1h, `×2` <1d, `×0.5` <1w, `×0.25` older)

### Implementation
- **DB**: migration `002` converts the existing RFC 3339 `added_at` column to an INTEGER epoch and adds `total_play_count` / `last_played_at` columns
- **Lib**: `track_ops` surfaces the new columns; `frecency_score()` implements the recency weighting
- **Playback**: `sort_by_mode` gains MostPlayed / Recency / Frecency; `record_track_play` increments `total_play_count` and sets `last_played_at` on track start
- **TUI**: Sort popup extended with the three new criteria and their key bindings

### Depends on
- PR #746 (base criteria + `SortPlaylist` RPC)

Implements upstream issue #709.


lib/src/config/v2/tui/keys/mod.rs
```diff
@@ -783,6 +783,32 @@ pub struct KeysPlaylist {
...
+    /// Sort playlist by date added, oldest first (Alt+Shift+t)
+    pub sort_date_added_desc: KeyBinding,
+    /// Sort playlist by frecency score, highest first (Alt+b)
+    pub sort_frecency_asc: KeyBinding,
+    /// Sort playlist by frecency score, lowest first (Alt+Shift+b)
+    pub sort_frecency_desc: KeyBinding,
+    /// Sort playlist by duration, shortest first (Alt+d)
+    pub sort_duration_asc: KeyBinding,
+    /// Sort playlist by duration, longest first (Alt+Shift+d)
+    pub sort_duration_desc: KeyBinding,
```
hasezoey: This likely would be better to have in another struct like `Sort` so that in the config it could appear as `keys.playlist.sort`, WDYT?

lib/src/config/v2/tui/keys/mod.rs
```diff
@@ -1963,6 +2063,67 @@ mod v1_interop {
...
+                    sort_duration_asc: tuievents::KeyEvent::new(
+                        tuievents::Key::Char('d'),
+                        tuievents::KeyModifiers::ALT,
+                    )
+                    .into(),
+                    sort_duration_desc: tuievents::KeyEvent::new(
+                        tuievents::Key::Char('D'),
+                        tuievents::KeyModifiers::ALT | tuievents::KeyModifiers::SHIFT,
+                    )
+                    .into(),
```
hasezoey: Once again, this should use the defaults instead of re-defining the defaults.

playback/src/lib.rs
```diff
@@ -500,7 +502,19 @@ impl GeneralPlayer {
                 self.add_and_play(&track).await;
             };
             Handle::current().block_on(wait);
-            self.run_info.write().set_current_track(track);
+            self.run_info.write().set_current_track(track.clone());
+
+            // Bump play_count and last_played for local music tracks (frecency sort support).
+            if track.media_type() == termusiclib::track::MediaTypesSimple::Music
```
hasezoey:
```suggestion
            if track.media_type() == MediaTypesSimple::Music
```

playback/src/lib.rs
```diff
@@ -500,7 +502,19 @@ impl GeneralPlayer {
...
+                && let Some(path) = track.path()
+            {
+                let conn = self.db.get_connection();
+                let _ = track_ops::increment_play_count(&conn, path);
+                let now = std::time::SystemTime::now()
```
hasezoey:
```suggestion
                let now = SystemTime::now()
```

playback/src/lib.rs
```diff
@@ -500,7 +502,19 @@ impl GeneralPlayer {
...
+                    .duration_since(std::time::UNIX_EPOCH)
+                    .map_or(0, |d| d.as_secs().cast_signed());
+                let _ = track_ops::set_last_played(&conn, path, now);
+            }
```
hasezoey: This does not handle the enqueued path, which is the termusic default.

playback/src/lib.rs
```diff
@@ -500,7 +502,19 @@ impl GeneralPlayer {
                 self.add_and_play(&track).await;
             };
             Handle::current().block_on(wait);
-            self.run_info.write().set_current_track(track);
+            self.run_info.write().set_current_track(track.clone());
```
hasezoey: Unless i missed something, this can be moved after the next `if` to avoid the clone.

playback/src/playlist.rs
```diff
@@ -989,6 +991,101 @@ impl Playlist {
         ));
     }
 
+    /// Reorder the playlist by `criterion` + `direction` and emit `PlaylistShuffled`.
+    ///
+    /// Non-alphabetical ties break alphabetically by title (natural sort).
```
hasezoey: This reads weird, please reword. (are some quotes missing?)

playback/src/playlist.rs
```diff
@@ -989,6 +991,101 @@ impl Playlist {
...
+                                SortCriterion::Download => at.unwrap_or(i64::MIN) as f64,
+                                SortCriterion::Frecency => track_ops::frecency_score(pc, lp, now),
+                                _ => 0.0,
+                            }
+                        }
+                    };
+                    (t.clone(), primary, title)
+                })
+                .collect()
+        };
```
hasezoey: This has too much indentation. Move this to a separate function.
Also consider using a named struct instead of a unnamed tuple.

tui/src/ui/components/config_editor/key_combo.rs
```diff
@@ -2156,6 +2301,89 @@ impl Model {
...
+            Box::new(key_playlist_sort_duration_asc(self.config_tui.clone())),
+            Vec::new(),
+        )?;
+        self.app.remount(
+            Id::ConfigEditor(IdConfigEditor::KeyOther(
+                IdKeyOther::PlaylistSortDurationDesc,
+            )),
+            Box::new(key_playlist_sort_duration_desc(self.config_tui.clone())),
+            Vec::new(),
+        )?;
```
hasezoey, line 2173: Consider having a separate function for just those keys (to also avoid the now ignored clippy lint)

tui/src/ui/msg.rs
```diff
@@ -673,6 +685,11 @@ pub enum PLMsg {
     AddRandomAlbum,
     /// Start choosing random tracks to be added to the playlist
     AddRandomTracks,
+    /// Sort the playlist by criterion (proto `SortCriterion` i32) and direction (proto `SortDirection` i32)
+    Sort {
+        criterion: i32,
+        direction: i32,
```
hasezoey: This should make use of enums too, as we practically do with most other things.
Looking at magic numbers is not that fun.

As `TrackSortCriterion` and direction are both in lib, how about reusing those here?

tui/src/ui/tui_cmd.rs
```diff
@@ -38,6 +38,10 @@ pub enum PlaylistCmd {
     Clear,
     SwapTrack(PlaylistSwapTrack),
     Shuffle,
+    Sort {
+        criterion: i32,
+        direction: i32,
```
hasezoey: Same here, magic numbers are not fun to figure out.

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,37 @@ use crate::new_database::{
...
 
+/// Sort criterion for playlist ordering.
+///
+/// Each criterion has a natural (lowercase key) and a reversed (Shift+key) direction.
+/// Ties on any non-alphabetical criterion break alphabetically by title (natural sort).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum TrackSortCriterion {
+    /// Sort by title alphabetically (A→Z).
+    #[default]
+    Alphabetical,
```
hasezoey: Please rename this and anything related to `Alphanumeric` instead of `Alphabetical`.

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,37 @@ use crate::new_database::{
...
+    /// Sort by play count (most played first in Desc).
+    Frequency,
```
hasezoey: As this is all time, please rename this to something along the line of `Most Played`

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,37 @@ use crate::new_database::{
...
+    /// Sort by last played time (most recent first in Desc).
+    Recency,
+    /// Sort by date added / download time (newest first in Desc). Uses existing `added_at`.
+    Download,
```
hasezoey: `Download` makes sense for podcasts, but for everything else it does not (at least in my opinion).
Please rename to `First Added` or `First Seen` or something along those lines.

Also does it really make sense to put `Date Added` and `(podcast) Download Time` in the same sort? In my opinion those are 2 different things.
If `Download time` is really for podcasts only, wouldnt it make more sense to sort by `Episode publish date`?

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,37 @@ use crate::new_database::{
...
+    /// zoxide-style frecency score (highest first in Desc).
+    Frecency,
```
hasezoey: I dont really like this name, it looks / sound like a typo. Is this really what it is called?

lib/src/new_database/migrations/003.sql
hasezoey, line 1: This file seems to be completely unused.
And all the functions in here seem to also exist in `002.sql`.

lib/src/new_database/migrations/001.sql
hasezoey, line 1: At least from my understanding of SQL migrations, past migrations should **never** be modified again.

lib/src/new_database/migrations/002.sql
```diff
@@ -0,0 +1,11 @@
...
+
+-- play count: incremented once per distinct track start
+ALTER TABLE tracks ADD COLUMN play_count INTEGER NOT NULL DEFAULT 0;
+-- last time the track was started (unix epoch seconds), NULL if never played
+ALTER TABLE tracks ADD COLUMN last_played INTEGER;
+
+-- Reset added_at: this column was added with DATE type (RFC 3339 strings) but
+-- we now use INTEGER (unix epoch seconds). All existing tracks were imported at
+-- once, so just set them all to the current timestamp.
+UPDATE tracks SET added_at = CAST(strftime('%s', 'now') AS INTEGER);
```
hasezoey: Why the switch to INTEGER instead of DATE?

lib/src/new_database/migrations/002.sql
```diff
@@ -0,0 +1,11 @@
+--- SECTION: Frecency sort support
+
+-- play count: incremented once per distinct track start
+ALTER TABLE tracks ADD COLUMN play_count INTEGER NOT NULL DEFAULT 0;
```
hasezoey: Consider renaming this to `total_play_count`.
This would better convey what / how it counts.
For example if in the future we should add separate counts for "this month", "past week", "past year", etc.

lib/src/new_database/migrations/002.sql
```diff
@@ -0,0 +1,11 @@
...
+-- last time the track was started (unix epoch seconds), NULL if never played
+ALTER TABLE tracks ADD COLUMN last_played INTEGER;
```
hasezoey: Rename to `last_played_at`.

lib/src/new_database/track_ops.rs
```diff
@@ -581,6 +632,85 @@ pub fn all_distinct_directories(conn: &Connection) -> Result<Vec<String>> {
...
+pub fn frecency_score(play_count: i64, last_played: Option<i64>, now: i64) -> f64 {
+    let rank = (play_count.max(0) as u64) as f64;
+    if rank == 0.0 {
+        return 0.0;
+    }
+    let multiplier = match last_played {
+        None => 0.25,
+        Some(lp) => {
+            let elapsed = now.saturating_sub(lp);
+            if elapsed < 3600 {
```
hasezoey: For ease of viewing/skimming, please add comments explaining what the static seconds represent.

lib/src/new_database/track_ops.rs
```diff
@@ -581,6 +632,85 @@ pub fn all_distinct_directories(conn: &Connection) -> Result<Vec<String>> {
...
+            } else if elapsed < 604_800 {
+                0.5
+            } else {
+                0.25
+            }
+        }
+    };
+    rank * multiplier
+}
+
```
hasezoey: I dont think this is the right file for this. It does not seem to be related to any database functionality.
Maybe consider moving to `utils` instead.

lib/src/new_database/track_ops.rs
```diff
@@ -1446,4 +1582,40 @@ mod tests {
...
+        assert!((frecency_score(1, Some(now - 100_000), now) - 0.5).abs() < f64::EPSILON);
+
+        // played > 1 week ago → × 0.25
+        assert!((frecency_score(1, Some(now - 700_000), now) - 0.25).abs() < f64::EPSILON);
+
+        // never played → × 0.25
+        assert!((frecency_score(1, None, now) - 0.25).abs() < f64::EPSILON);
+
+        // multiple plays
+        assert!((frecency_score(3, Some(now - 100_000), now) - 1.5).abs() < f64::EPSILON);
```
hasezoey: Do those tests not just test that *a value* is returned instead of *a specific value*? Am i misunderstanding those tests?

hasezoey (changes requested):
Thank you for trying to implement these changes, but there is a lot to go over with this one.

The biggest issue, in my opinion, is that this is **one single commit**. This should be split up into specific parts (eg. server side impl; protobuf impl, tui impl). This is required by our [CONTRIBUTING: Size of the commits](https://github.com/tramhao/termusic/blob/master/CONTRIBUTING.md#size-of-the-commits).

Additionally, please split this PR into 2: one for the base addition of the sorts for existing fields and one PR for new sorts that require database changes.

This is likely gonna be a lot more back-and-forth than the previous PRs.

Now as for the code itself, i dont quite like how many keys are added this time, are you sorting *that often* that you dont just want a *single* key to open a menu to select the sort to be done?
Additionally, it would be great to even *have* a popup sort menu, instead of only being able to figure it out via key shortcuts.

RG-IL:
"Frecency" is an established term in information retrieval — a portmanteau of "frequency" + "recency." It is used by:

- **zoxide** (the smart `cd` replacement): the entire ranking algorithm is called frecency, using the exact same bucket multipliers (x4 for <1hr, x2 for <1day, x0.5 for <1week, x0.25 for older). https://github.com/ajeetdsouza/zoxide/wiki/Algorithm
- **Mozilla Firefox**: the Places/History URL ranking system is called Frecency — "a term derived from frequency and recency, its scope is to provide a ranking algorithm that gives importance both to how often a page is accessed and when it was last visited." https://firefox-source-docs.mozilla.org/browser/urlbar/ranking-legacy.html

I use zoxide daily and it is battle-tested. I would prefer to keep this name for consistency with the ecosystem.

hasezoey:
> I use zoxide daily and it is battle-tested. I would prefer to keep this name for consistency with the ecosystem.

No problem, i just had not encountered that term before, and on first look it looked like a typo, now i know it is not.
For future readers (and likely myself), could you add some kind of note that this is not a typo (and/or the text `"frequency" + "recency"`)?

RG-IL:
All review points addressed:

- 12 keybindings replaced with `,` key → popup menu with key hints (ascending/lowercase, descending/uppercase)
- Config simplified to single `keys.playlist.sort` field
- Enums renamed: Alphabetical→Alphanumeric, Frequency→MostPlayed, Download→FirstAdded
- Column renames: `play_count`→`total_play_count`, `last_played`→`last_played_at`
- DB migrations cleaned (001.sql reverted, 003.sql deleted)
- `frecency_score` uses `FrecencyTimeBucket` enum for magic numbers + doc comment noting zoxide etymology
- Code quality: `ScoredTrack` struct, extracted scoring/sorting functions, typed enums in TUI messages
- Playback: `track.clone()` moved after guard, play count bumped for enqueued path
- Config editor consolidated to single key-combo widget
- Sort docstring rewritten

**Question on PR structure:** You originally asked for 2 PRs — one for base sorts (alphanumeric, duration) and one for new DB-dependent sorts (most played, recency, first added, frecency). However, the popup menu lists all 6 options together, and splitting means PR1 ships a popup showing 4 broken options until PR2 lands. Would you prefer:

(a) 2 PRs as originally requested (infrastructure + 2 working sorts, then DB + 4 remaining sorts)
(b) 1 PR with clean commit history: infrastructure commits → DB migration → final wiring

RG-IL:
@hasezoey  this commit is Just temporarily for you to decide which commit path and PR path to take.

I've tested on my system, and all of the sorting works perfectly.

.cortexkit/.gitignore
hasezoey, line 1: Why does this exist? Shouldnt this be in the top-level gitignore?

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,40 @@ use crate::new_database::{
...
+    /// Sort by title alphanumerically (A→Z).
+    #[default]
+    Alphanumeric,
+    /// Sort by total play count (most played first in Desc).
+    MostPlayed,
+    /// Sort by last played time (most recent first in Desc).
+    Recency,
+    /// Sort by date first seen in the library (newest first in Desc). Uses `added_at`.
+    FirstAdded,
+    /// Hybrid of **freq**uency and rec**ency** — zoxide-style score (highest first in Desc).
```
hasezoey:
```suggestion
    /// Hybrid of **fre**quency and re**cency** — zoxide-style score (highest first in Desc).
```

lib/src/new_database/migrations/002.sql
```diff
@@ -0,0 +1,15 @@
...
+ALTER TABLE tracks ADD COLUMN total_play_count INTEGER NOT NULL DEFAULT 0;
+-- Unix epoch seconds of the last time the track was started. NULL if never played.
+ALTER TABLE tracks ADD COLUMN last_played_at INTEGER;
+
+-- added_at was originally defined as DATE (RFC 3339 strings) but we now need
+-- INTEGER (unix epoch seconds) for numeric sorting. SQLite does not support
+-- altering a column type in place, so drop and re-create it. All existing tracks
+-- were imported at once, so backfilling with the current timestamp is correct.
+ALTER TABLE tracks ADD COLUMN added_at_new INTEGER;
+UPDATE tracks SET added_at_new = CAST(strftime('%s', 'now') AS INTEGER);
```
hasezoey: I might not know too much about SQL, but this does not look like it actually tries to import from the old `added_at` anymore and instead sets it to the current time the migration runs.

lib/src/config/v2/tui/keys/mod.rs
```diff
@@ -783,6 +783,10 @@ pub struct KeysPlaylist {
     /// previously known as `cmus_lqueue`
     // NOTE: currently this can be somewhat broken sometimes, cause unknown
     pub add_random_album: KeyBinding,
+
+    // --- Playlist sort keys (all use ALT modifier to avoid conflicts) ---
```
hasezoey:
```suggestion
```

This no longer looks relevant

lib/src/config/v2/tui/keys/mod.rs
```diff
@@ -1963,6 +1975,12 @@ mod v1_interop {
                     swap_down: value.playlist_swap_down.into(),
                     add_random_songs: value.playlist_add_random_tracks.into(),
                     add_random_album: value.playlist_add_random_album.into(),
+                    // Sort key: no v1 equivalent, use comma default
+                    sort: tuievents::KeyEvent::new(
+                        tuievents::Key::Char(','),
+                        tuievents::KeyModifiers::empty(),
+                    )
+                    .into(),
```
hasezoey: Once again this should use the default instead of re-defining.

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,40 @@ use crate::new_database::{
 
 use super::Integer;
 
+pub use crate::utils::frecency_score;
+
+/// Sort criterion for playlist ordering.
+///
+/// Each criterion has a natural (lowercase key) and a reversed (Shift+key) direction.
```
hasezoey:
```suggestion
```

This is no longer relevant (and was not in the first place)

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,40 @@ use crate::new_database::{
...
+/// Ties on any non-alphanumeric criterion break by track title (natural sort).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum TrackSortCriterion {
+    /// Sort by title alphanumerically (A→Z).
```
hasezoey:
```suggestion
    /// Sort by title alphanumerically (A→Z; 1→100+).
```

lib/src/new_database/track_ops.rs
```diff
@@ -16,6 +16,40 @@ use crate::new_database::{
...
+    #[default]
+    Alphanumeric,
+    /// Sort by total play count (most played first in Desc).
+    MostPlayed,
+    /// Sort by last played time (most recent first in Desc).
+    Recency,
+    /// Sort by date first seen in the library (newest first in Desc). Uses `added_at`.
```
hasezoey:
```suggestion
    /// Sort by date first seen in the library (newest first in Desc). Uses `added_at` for tracks and `pubdate` for episodes.
```

lib/src/new_database/track_ops.rs
```diff
@@ -60,6 +94,12 @@ pub struct TrackRead {
     // Direct data on `tracks`
     pub duration: Option<Duration>,
     pub last_position: Option<Duration>,
+    /// Total number of times this track has been started (incremented once per start).
+    pub total_play_count: i64,
```
hasezoey: As it never can actually go in the negative, shouldnt this be unsigned?
(i know sqlite itself does not have such a type, but this is not the sql database anymore)

lib/src/new_database/track_ops.rs
```diff
@@ -581,6 +635,81 @@ pub fn all_distinct_directories(conn: &Connection) -> Result<Vec<String>> {
...
+        conn.execute(
+            "INSERT OR IGNORE INTO tracks (file_dir, file_stem, file_ext, total_play_count)
+            VALUES (:file_dir, :file_stem, :file_ext, 1);",
+            named_params! {
+                ":file_dir": file_dir.as_ref(),
+                ":file_stem": file_stem.as_ref(),
+                ":file_ext": file_ext.as_ref(),
+            },
+        )?;
+    }
```
hasezoey: I dont think this function should insert a track; that is handled specifically by other functions.
Also the database is intended to only contain values from the set music libraries, and not track ones that are not from there (though currently the TUI indexes all the paths that are visited in the filebrowser)

lib/src/new_database/track_ops.rs
```diff
@@ -581,6 +635,81 @@ pub fn all_distinct_directories(conn: &Connection) -> Result<Vec<String>> {
...
+            "INSERT OR IGNORE INTO tracks (file_dir, file_stem, file_ext, last_played_at)
+            VALUES (:file_dir, :file_stem, :file_ext, :timestamp);",
+            named_params! {
+                ":timestamp": timestamp,
+                ":file_dir": file_dir.as_ref(),
+                ":file_stem": file_stem.as_ref(),
+                ":file_ext": file_ext.as_ref(),
+            },
+        )?;
+    }
```
hasezoey: Once again, i dont think this function should insert a new track

lib/src/utils.rs
```diff
@@ -219,6 +219,74 @@ pub fn random_ascii(len: usize) -> String {
         .collect()
 }
 
+/// "Frecency" score — a hybrid of **freq**uency and rec**ency**, popularized
```
hasezoey:
```suggestion
/// "Frecency" score — a hybrid of **fre**quency and re**cency**, popularized
```

lib/src/utils.rs
```diff
@@ -431,6 +499,34 @@ mod tests {
...
+        assert!((frecency_score(1, Some(now - 100_000), now) - 0.5).abs() < eps);
+
+        // 1 play, ≥ 1 week ago → rank 1 × 0.25 = 0.25
+        assert!((frecency_score(1, Some(now - 700_000), now) - 0.25).abs() < eps);
+
+        // 1 play, never played → rank 1 × 0.25 = 0.25
+        assert!((frecency_score(1, None, now) - 0.25).abs() < eps);
+
+        // 3 plays, < 1 week ago → rank 3 × 0.5 = 1.5
+        assert!((frecency_score(3, Some(now - 100_000), now) - 1.5).abs() < eps);
```
hasezoey, line 527: Once again, are these not just testing that a float is below a certain value instead of more exactly testing a value?

playback/src/lib.rs
```diff
@@ -478,6 +480,18 @@ impl GeneralPlayer {
...
+                if track.media_type() == MediaTypesSimple::Music
+                    && let Some(path) = track.path()
+                {
+                    let conn = self.db.get_connection();
+                    let _ = track_ops::increment_total_play_count(&conn, path);
+                    let now = SystemTime::now()
+                        .duration_since(UNIX_EPOCH)
+                        .map_or(0, |d| d.as_secs().cast_signed());
+                    let _ = track_ops::set_last_played_at(&conn, path, now);
+                }
```
hasezoey: As this is now practically literally copied, why not make this a function?

playback/src/playlist.rs
```diff
@@ -989,6 +992,51 @@ impl Playlist {
...
+
+        let mut scored: Vec<ScoredTrack> = self
+            .tracks
+            .iter()
+            .map(|t| score_track(t, criterion, db, now))
+            .collect();
+
+        sort_scored(&mut scored, criterion, direction);
+
+        self.tracks = scored.into_iter().map(|s| s.track).collect();
```
hasezoey: This is looping over the array 3 times and 2 times creating a new vec, couldnt this be reduced with something like `Vec::sort_by_key` / `::sort_by_cached_key`?

playback/src/playlist.rs
```diff
@@ -1109,6 +1157,92 @@ impl Playlist {
...
+
+            // File mtime is the best source for "when was this track first seen" —
+            // the DB column was bulk-set to a single timestamp during migration.
+            let file_mtime = track.path().and_then(|p| {
+                std::fs::metadata(p)
+                    .ok()
+                    .and_then(|m| m.modified().ok())
+                    .and_then(|t| t.duration_since(std::time::UNIX_EPOCH).ok())
+                    .map(|d| d.as_secs().cast_signed())
+            });
```
hasezoey: I do not like that a *sort* accesses the filesystem and requests metadata.
This should only look at the database values, unless there is a specific sort like `File MTime`.

playback/src/playlist.rs
```diff
@@ -1109,6 +1157,92 @@ impl Playlist {
...
+
+            match criterion {
+                SortCriterion::MostPlayed => pc as f64,
+                SortCriterion::Recency => lp.unwrap_or(file_mtime.unwrap_or(i64::MIN)) as f64,
+                SortCriterion::FirstAdded => file_mtime.unwrap_or(i64::MIN) as f64,
+                SortCriterion::Frecency => {
+                    // Use file mtime as a stand-in for last_played_at when the
+                    // track has never been played — gives newer files higher scores.
+                    let effective_lp = lp.or(file_mtime);
+                    track_ops::frecency_score(pc, effective_lp, now)
```
hasezoey: I do not know how you got it to compile (in my quick skim), but that function is no longer in that location, is it?

playback/src/playlist.rs
```diff
@@ -1109,6 +1157,92 @@ impl Playlist {
...
+    }
+}
+
+/// Sort a scored track list in-place according to `criterion` + `direction`.
+fn sort_scored(scored: &mut [ScoredTrack], criterion: SortCriterion, direction: SortDirection) {
+    let sign = if direction == SortDirection::Desc {
+        -1.0
+    } else {
+        1.0
+    };
```
hasezoey: If this is only necessary in a specific path, it should be moved there

tui/src/ui/components/config_editor/key_combo.rs
```diff
@@ -1739,6 +1740,15 @@ fn key_playlist_add_random_tracks(config: SharedTuiSettings) -> KEModifierSelect
     )
 }
 
+#[inline]
+fn key_playlist_sort(config: SharedTuiSettings) -> KEModifierSelect {
+    KEModifierSelect::new(
+        " Playlist Sort ",
```
hasezoey:
```suggestion
        " Playlist Sort Popup ",
```

tui/src/ui/components/config_editor/key_combo.rs
```diff
@@ -2101,6 +2111,7 @@ impl Model {
     }
 
     /// Mount / Remount the Config-Editor's Third Page, the Playlist key-combos
+    #[allow(clippy::too_many_lines)]
```
hasezoey: Is this still necessary?

tui/src/ui/components/popups/sort.rs
```diff
@@ -0,0 +1,176 @@
...
+                )
+                .inactive(Style::new().bg(config.settings.theme.library_background()))
+                .foreground(config.settings.theme.fallback_foreground())
+                .background(config.settings.theme.fallback_background())
+                .highlight_style(
+                    CommonHighlight::default()
+                        .style
+                        .fg(config.settings.theme.fallback_highlight()),
+                )
+                .scroll(false)
```
hasezoey: Should it really not be scrollable (select elements with ARROW UP/DOWN + ENTER)?

tui/src/ui/components/popups/sort.rs
```diff
@@ -0,0 +1,176 @@
...
+        );
+        self.update_photo().ok();
+        assert!(self.app.active(&Id::SortPopup).is_ok());
+    }
+
+    pub fn umount_sort_popup(&mut self) {
+        if self.app.mounted(&Id::SortPopup) {
+            self.app.umount(&Id::SortPopup).ok();
+        }
+    }
```
hasezoey: This can be done unconditionally.

tui/src/ui/components/popups/sort.rs
```diff
@@ -0,0 +1,176 @@
...
+        None
+    }
+}
+
+impl AppComponent<Msg, UserEvent> for SortPopup {
+    fn on(&mut self, ev: &Event<UserEvent>) -> Option<Msg> {
+        match ev {
+            Event::Keyboard(KeyEvent {
+                code: Key::Char(ch),
+                modifiers: KeyModifiers::NONE | KeyModifiers::SHIFT,
```
hasezoey: I dont think this works as you would expect it to as this is a *binary* OR instead of a *logical* OR of multiple chocie.

tui/src/ui/components/popups/sort.rs
```diff
@@ -0,0 +1,176 @@
...
+            }) => {
+                if let Some((criterion, direction)) = Self::match_key(*ch) {
+                    return Some(Msg::SortPopup(SortPopupMsg::Selected(criterion, direction)));
+                }
+                Some(Msg::SortPopup(SortPopupMsg::Close))
```
hasezoey, line 225: Are you sure you want to always close the popup on wrong keys?

playback/src/playlist.rs
```diff
@@ -989,6 +992,51 @@ impl Playlist {
         ));
     }
 
+    /// Reorder the playlist by `criterion` + `direction` and emit `PlaylistShuffled`.
+    ///
+    /// Ties on non-alphanumeric criteria break by track title (natural sort).
```
hasezoey:
```suggestion
    /// Ties on non-alphanumeric criteria are broken by track title (natural sort).
```

hasezoey (changes requested):
> Question on PR structure: You originally asked for 2 PRs — one for base sorts (alphanumeric, duration) and one for new DB-dependent sorts (most played, recency, first added, frecency). However, the popup menu lists all 6 options together, and splitting means PR1 ships a popup showing 4 broken options until PR2 lands. Would you prefer:
> 
> (a) 2 PRs as originally requested (infrastructure + 2 working sorts, then DB + 4 remaining sorts)
> (b) 1 PR with clean commit history: infrastructure commits → DB migration → final wiring

I do not know why it would show broken entries in the popup if those entries do not exist at that time? (Or said differently, PR 2 should add the new options)

Please go with the 2 PR way, this makes review and reasoning on changes easier.

> one for base sorts (alphanumeric, duration) and one for new DB-dependent sorts (most played, recency, first added, frecency).

Shouldnt `first added` (`added_at` & `pubdate`) not be in the first PR as database modification is not required?

---

Please resolve conversations (sub threads) if they are unambiguously resolved.

Additionally, please answer my question about `Why the switch to INTEGER instead of DATE?`.

tui/src/ui/components/popups/sort.rs
RG-IL: I think it's better as a key hint because it makes it faster and, for me, more reliable. 

This is the same approach as the Yazi File Explorer TUI.

hasezoey: Yes it might make it faster for some users and that functionality can stay, but it would make it inconsistent with anything else in termusic tui right now not being able to be selected via ARROW keys and ENTER.

RG-IL: maybe i will make it a option to change between the ARROW and ENTER and the key hint, and make the ARROW keys and ENTER the deafult?

hasezoey: > maybe i will make it a option to change between the ARROW and ENTER and the key hint, and make the ARROW keys and ENTER the deafult?

I dont know why you would need a option for that. From my POV both can co-exists enabled at the same time. (i guess unless the vim navigation keys are bound to those keys...)

EDIT: if you really see another conflict, please change it to use the current termusic way, and then later we can add you faster way.

RG-IL: You're actually right. They can coexist together. Thank you. I didn't think about it.

RG-IL:
SQLite has no dedicated DATE storage type. DATE in a CREATE TABLE gets NUMERIC affinity — not a special date format. Dates must be stored as TEXT (ISO 8601), INTEGER (unix epoch), or REAL (Julian day) by the application.
I chose INTEGER because:
1. Frecency arithmetic — the score computes now - last_played_at to bucket into time ranges (<1hr, <1day, <1week, older). On integers that's a subtraction. On text you'd need julianday() or strftime('%s', ...) conversions on every sort call.
2. Numeric sort — ORDER BY added_at DESC on integers is a single CPU comparison. Text works (ISO 8601 sorts lexicographically correctly) but is slower.
3. Standard recommendation — SQLite's own docs list INTEGER as unix epoch seconds as the first-class numeric date format. Most performance-oriented applications default to it.
TEXT dates would also work fine for a music library. The real driver was that our Rust sort code was written around integer timestamps — converting the other way (keeping TEXT, wrapping every comparison in julianday()) would have been more friction for no benefit.

RG-IL:
@hasezoey I'm think the changes you made to the playlist code made it so the sorting currently doesn't work, also i will be offline until next Monday. 

So i will close both PRs and will make new ones with fixes and cleanups

RG-IL:
I rebased this onto the master for it be updated, then it stopped working so I thought it was because of the changes.

Turns out it's just because I used my custom fork and when I tested. I forgot to relaunch the server from scratch.

I will be offline until monday so i thought to close it as i thought it did not work and i will not have rhe time ubtil monday.

Now i understand that it does in fact work i will open it again

RG-IL:
Hi , im sorry for the confusion, i will explain the current state of both PRs (#746 and #743) for more clarity.
1. I've reorganized the commits so they will be like @hasezoey wanted (lib, tui, server etc.)
2. Both PRs work perfectly
3. Both PRs are ready for review 
4. I won't be able to change the code until Monday as i will not have my computer

hasezoey:
Just as a FYI, i wont review this PR until #746 is done and merged.

RG-IL:
@hasezoey I fixed merge conflicts

lib/src/new_database/track_insert.rs
```diff
@@ -259,7 +259,7 @@ impl InsertTrack<'_> {
                 ":file_ext": self.file_ext,
                 ":duration": duration,
                 ":last_position": last_position,
-                ":added_at": &now,
+                ":added_at": now,
```
RG-IL: i will change back to old as it doesnt change anything and will only add changes for reviewers to look at

lib/src/config/v2/tui/keys/mod.rs
RG-IL, line 2173: i will update to use deafult

server/src/server.rs
```diff
@@ -534,6 +534,7 @@ fn player_loop(
                 player.playlist.write().shuffle();
             }
             PlayerCmd::PlaylistSort(criterion, direction) => {
+                info!("sorting playlist by {criterion:?} {direction:?}");
```
RG-IL: I will do that it won’t log, as it didn’t in PR1, and it shouldn’t be different here.

.gitignore
```diff
@@ -5,4 +5,4 @@
 .cargo
 # VSCode relation configuration
 .vscode
-.cortexkit/
+.cortexkit
```
RG-IL: will do like the master so it will not have to be a change to review

RG-IL (commented):
I reviewed myself.

lib/src/new_database/migrate.rs
```diff
@@ -55,6 +55,15 @@ fn apply_migrations(conn: &Connection, mut user_version: u32) -> Result<()> {
         set_db_created_with(conn)?;
     }
 
+    if user_version == 1 {
+        // Add total_play_count and last_played_at columns for sort support (MostPlayed, Recency, FirstAdded, Frecency)
```
hasezoey:
```suggestion
        // Add total_play_count and last_played_at columns for sort support (MostPlayed, Recency, Frecency) plus `added_at` column type change
```

lib/src/new_database/track_ops.rs
```diff
@@ -60,8 +60,12 @@ pub struct TrackRead {
...
     pub duration: Option<Duration>,
     pub last_position: Option<Duration>,
-    /// Date the track was first added to the library (used for the "first added" sort).
-    pub added_at: Option<chrono::DateTime<chrono::Utc>>,
+    /// Total number of times this track has been started (incremented once per start).
+    pub total_play_count: u64,
+    /// Unix epoch seconds of the last time this track was started, or `NULL` if never played.
+    pub last_played_at: Option<i64>,
+    /// Unix epoch seconds of when this track was added to the library, or `NULL`.
+    pub added_at: Option<i64>,
```
hasezoey: I am pretty sure Unix epoch seconds cannot be negative, so it would be more correct to use a unsigned type, even if that requires some type conversions after fetching.

lib/src/new_database/track_ops.rs
```diff
@@ -500,12 +516,10 @@ fn common_row_to_trackread(conn: &Connection, row: &Row<'_>) -> TrackRead {
...
     };
 
-    let added_at: Option<String> = row.get("added_at").unwrap_or(None);
-    let added_at = added_at.and_then(|s| {
-        chrono::DateTime::parse_from_rfc3339(&s)
-            .ok()
-            .map(|dt| dt.with_timezone(&chrono::Utc))
-    });
+    #[allow(clippy::cast_sign_loss)] // play counts are always ≥ 0
+    let total_play_count: u64 = row.get::<_, i64>("total_play_count").unwrap_or(0) as u64;
```
hasezoey:
```suggestion
    let total_play_count: u64 = row.get::<_, i64>("total_play_count").unwrap_or(0).max(0).cast_unsigned();
```

Our MSRV is 1.90, allowing us to use [`i64::cast_unsigned`](https://doc.rust-lang.org/std/primitive.i64.html#method.cast_unsigned), which should avoid the clippy warning and the "silent" warning loss when using `as` if the types involved change.

The `max` is just added in case something weird happens and a negative number actually does get stored (for example a mistaken user action on the database).

lib/src/utils.rs
```diff
@@ -219,6 +219,74 @@ pub fn random_ascii(len: usize) -> String {
...
+///   < 1 day   → × 2.0
+///   < 1 week  → × 0.5
+///   else      → × 0.25
+#[must_use]
+#[allow(clippy::cast_precision_loss)]
+pub fn frecency_score(total_play_count: u64, last_played_at: Option<i64>, now: i64) -> f64 {
+    let rank = total_play_count as f64;
+    if rank == 0.0 {
+        return 0.0;
+    }
```
hasezoey: This could be done before the cast, to avoid float weirdness.

lib/src/utils.rs
```diff
@@ -431,6 +499,34 @@ mod tests {
         assert_eq!(iter.next(), None);
     }
 
+    #[test]
+    #[allow(clippy::float_cmp, clippy::float_equality_without_abs)]
```
hasezoey: I know i kinda complained earlier about "if the test actually does something", but we shouldnt rely on exact float values (even just multiplies) due to platform differences here.
At least that is my understanding of floats in rust.

playback/src/lib.rs
```diff
@@ -531,6 +537,27 @@ impl GeneralPlayer {
         }));
     }
 
+    /// Record that a track was started — bumps play count and last-played timestamp.
+    fn record_track_play(&self, track: &Track) {
```
hasezoey, line 550: This is something that can be done in a follow-up, but in my opinion those 2 values should not *always* be set / incremented when starting to play, but after a threshold, ex. 5s, because skipping through some songs shouldnt mark them as having been "played".
Also should resumes (starting at a time fetched from the database) actually count as another play?

playback/src/playlist.rs
```diff
@@ -1009,16 +1013,23 @@ impl Playlist {
...
+            0
+        };
 
-        // take the original vec as this allows us to use the existing "Track"s already without cloning
-        let orig_track = std::mem::take(&mut self.tracks);
-
-        // Another vec is necessary as "sort_by_*_key" require that the returned key
-        // is not tied to a lifetime, which it in this case it would require
-        // to match against the track's title.
-        let mut scored: Vec<ScoredTrack> = orig_track
```
hasezoey, line 1019: Why was this seemingly reverted?

playback/src/playlist.rs
```diff
@@ -1171,17 +1182,35 @@ struct ScoredTrack {
...
+            let (pc, lp, added) = tr.as_ref().map_or((0, None, None), |x| {
+                (x.total_play_count, x.last_played_at, x.added_at)
+            });
+
+            match criterion {
+                SortCriterion::MostPlayed => pc as f64,
+                SortCriterion::Recency => lp.map_or(f64::MIN, |v| v as f64),
+                SortCriterion::FirstAdded => added.map_or(f64::MIN, |v| v as f64),
+                SortCriterion::Frecency => frecency_score(pc, lp, now),
+                _ => 0.0,
```
hasezoey:
```suggestion
                _ => unreachable!(),
```

tui/src/ui/components/popups/sort.rs
```diff
@@ -34,6 +34,22 @@ const SORT_HINTS: &[SortHint] = &[
         asc_description: "Filename A\u{2192}Z",
         desc_description: "Filename Z\u{2192}A",
     },
+    SortHint {
+        ascending_key: 'f',
+        descending_key: 'F',
```
hasezoey: Wouldnt key `m` be more intuitive with the current name of `Most Played`?

tui/src/ui/components/popups/sort.rs
```diff
@@ -34,6 +34,22 @@ const SORT_HINTS: &[SortHint] = &[
...
+        criterion: SortCriterion::MostPlayed,
+        label: "Most Played",
+        asc_description: "Play count least\u{2192}most",
+        desc_description: "Play count most\u{2192}least",
+    },
+    SortHint {
+        ascending_key: 'c',
+        descending_key: 'C',
```
hasezoey: Here the same, wouldnt key `r` be more intuitive here for `Recency`?

hasezoey (changes requested):
This split PR makes it a lot easier to review.

Some questions and styling suggestions. Not actually tested to run yet (maybe tomorrow).
Otherwise looks good.

playback/src/lib.rs
RG-IL: You're right. I will fix it.

tui/src/ui/components/popups/sort.rs
RG-IL: I thought of frequency, m for most played is a good option too.

tui/src/ui/components/popups/sort.rs
RG-IL: At the start, when there wasn't a popup, I couldn't use R as it was reserved to be for the random sort.

playback/src/playlist.rs
RG-IL: The original mem::take + into_iter pattern avoided cloning by consuming the tracks vec, but our score_track takes &Track — it needs to borrow the track to call .path() and .duration() for DB lookups. Switching to into_iter would mean score_track takes Track by value, which would require passing db as &Database instead of borrowing it through the track's lifetime, and reworking the ScoredTrack construction. The clone here is per-track and only during sort operations, so the simpler iter().map() pattern seemed worth the trade-off.

hasezoey: I dont understand your argument, `score_track` *already* takes `&Database` and from my (dry) understanding i also dont see why `Track` cannot be taken anymore to avoid the clone.
Also `&Database` is in no way tied to `&Track` from what i can tell.

RG-IL: You are right. I will fix it.

playback/src/lib.rs
```diff
@@ -480,6 +493,9 @@ impl GeneralPlayer {
             };
             if enqueued_track.as_track_source() == track.as_track_source() {
                 info!("Starting seamless Track {track:#?}");
+
+                self.record_track_play(track);
```
hasezoey:
```suggestion
```

From what i can tell, that is already handled by the `record_track_play` at the start of the function

playback/src/lib.rs
```diff
@@ -512,6 +529,13 @@ impl GeneralPlayer {
             info!("Stopping due to not having a next track");
             drop(playlist);
             drop(run_info);
+            if self
+                .position()
+                .is_some_and(|pos| pos.as_secs() >= PLAY_THRESHOLD_SECS)
+                && let Some(current) = self.run_info.read().current_track()
+            {
+                self.record_track_play(current);
+            }
```
hasezoey:
```suggestion
```

From what i can tell, that is already handled by the `record_track_play` at the start of the function

lib/src/new_database/track_ops.rs
```diff
@@ -592,6 +619,51 @@ pub fn all_distinct_directories(conn: &Connection) -> Result<Vec<String>> {
...
+        },
+    )?;
+
+    Ok(())
+}
+
+/// Set `last_played_at` to the given unix epoch timestamp for the given track.
+///
+/// The track is identified by its file path. If the track is not found, this is a no-op.
+pub fn set_last_played_at(conn: &Connection, track: &Path, timestamp: i64) -> Result<()> {
```
hasezoey: Still takes a `i64` instead of a `u64`. Likely should match what `TrackRead` uses.

hasezoey (changes requested):
Looks good and works, some minor style things.

As this is likely the final time for this PR, could you cleanup the commits again too?

playback/src/lib.rs
```diff
@@ -453,6 +457,15 @@ impl GeneralPlayer {
...
     /// if `current_track_index` in playlist is above u32
     pub fn start_play(&mut self, from_eos: bool) {
+        // Record the outgoing track's play count if it played past the threshold.
+        if self
+            .position()
+            .is_some_and(|pos| pos.as_secs() >= PLAY_THRESHOLD_SECS)
+            && let Some(current) = self.run_info.read().current_track()
+        {
+            self.record_track_play(current);
+        }
```
hasezoey, line 467: Personally, i would add this to the `Tick` handler so that the play gets recorded even if the user quits later or just goes into state `Stopped` (which *should* clear the active track). Even if this would require some kind of flag that it has been recorded.
Though that can be done later.

hasezoey (approved):

PR #745, opened 2026-07-15
Change TUI Playlist Component to be a custom one

hasezoey:
This PR changes the TUI `Playlist` Component to be a custom one with a lot more customization options.
This is also prep-work for #493 on the TUI side (as we need direct access to tracks instead of a full string table).

Due to this re-write in a custom component, the following behavioral changes are done:
- Page Up/Down actually corresponds to the last drawn size now
- Artist & Title columns have been swapped (for a `Duration`, `Title`, `Artist`, `Album` order)
- Highlight color has been removed from being used for any column (previously used for `Artist`)
 - The `Title` column remains BOLD
- When in loop-mode `Playlist` or `PlaylistOnce`, past elements are now `LightBlack`
- The "Current Track" indicator is now in the same column as the "Current Highlight" symbol

Note that i plan to move this new Widget + Mock to be its own crate again, similar to what i did with `tuirealm-orx-tree`.

Simple screenshot comparison:

before
after


hasezoey:
Ubuntu or Github seems to be having network issues currently, failing to install the required packages and so not even running our tests currently.

PR #746, opened 2026-07-15
feat: playlist sort — base criteria (alphabetical, date-added, duration)

RG-IL:
This is **part 1 of 2** of the playlist sort feature. The DB-dependent criteria (most-played, recency, frecency) and their migration live in the follow-up PR #743.

Adds the ability to reorder the current playlist by a chosen criterion with ASC/DESC, via a new **Sort** popup. This PR covers the **base** criteria that need no database schema change:

- **Alphabetical** (`a` / `A`) — A–Z / Z–A, natural sort
- **Date-added** (`t` / `T`) — oldest / newest (`added_at`, read from its existing RFC 3339 value)
- **Duration** (`d` / `D`) — shortest / longest

### Implementation
- **Proto**: `SortCriterion` (Alphanumeric, FirstAdded, Duration) + `SortDirection` enums, `SortPlaylist` RPC
- **Server**: `sort_playlist` handler translates the request into a `PlayerCmd`; reuses the existing `PlaylistShuffled` event so the TUI reloads with no new plumbing
- **Playback**: `sort_by_mode()` scores tracks (duration as seconds; `added_at` RFC 3339 → epoch); ties fall back to title natural-sort
- **Lib**: `added_at` is already stored as RFC 3339 and is surfaced on `TrackRead` (no migration needed)
- **TUI**: Sort popup lists the criteria, maps keys to `SortCriterion`/`SortDirection`, sends `SortPlaylist`; keybinding + help + config-editor wiring

### Notes
- No DB migration in this PR — `added_at` is read as-is. The follow-up PR #743 adds the migration + DB-dependent criteria.
- Implements upstream issue #709.

PR #748, opened 2026-07-21
fix(playback,server): reset next_track_index after shuffle

RG-IL:
After `shuffle()` (and later `sort_by_mode()`), `next_track_index` was left stale — pointing to the wrong track in the now-reordered vec. This caused the gapless pipeline to potentially enqueue the wrong next track.

Clearing `run_info.enqueued` alongside ensures the backend does not hold a stale pre-fetched track either.

Relevant discussion: https://github.com/tramhao/termusic/pull/746#issuecomment-5028150578

### DIFF
--- a/playback/src/playlist.rs
+++ b/playback/src/playlist.rs
@@ -982,6 +982,8 @@ impl Playlist {
             self.current_track_index = index;
         }
 
+        self.next_track_index = None;
+
         self.send_stream_ev_pl(UpdatePlaylistEvents::PlaylistShuffled(
             PlaylistShuffledInfo {
                 tracks: self.as_grpc_playlist_tracks().unwrap(),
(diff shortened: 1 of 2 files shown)


hasezoey (changes requested):
Looks good, but now please update for the merge of #746.

playback/src/playlist.rs
```diff
@@ -985,6 +985,8 @@ impl Playlist {
             self.current_track_index = index;
         }
 
+        self.next_track_index = None;
```
hasezoey, line 988: Should now also be in `sort_by_mode`, right?

RG-IL: yes you right

hasezoey (approved):
Looks good and can confirm this fixed the (rather low odds) issue.

PR #750, opened 2026-07-27
feat(cli): add podcast and playlist subcommands

### LINKED ISSUE #304: Bash/cron command to auto  refresh all podcast feeds
## Describe what you want
Is it possible to issue a bash command, (for example in crontab) in order to automaticaaly refresh podcast feeds?
- I.E. Using an existing command to access the termusic database; it seems not , according to the command termusic-server --help

Thanks,
John

RG-IL:
## Summary

Adds CLI subcommands for podcast and playlist management:

**Podcast subcommand** (`termusic podcast`):
- `termusic podcast import <FILE>` — import feeds from OPML
- `termusic podcast export <FILE>` — export feeds to OPML
- `termusic podcast refresh` — refresh all podcast feeds

**Playlist subcommand** (`termusic podcast`):
- `termusic playlist shuffle` — shuffle the playlist
- `termusic playlist sort [criterion] [-d direction]` — sort by `alphanumeric` (default), `first-added`, `duration`, `most-played`, `recency`, or `frecency`
- `termusic playlist cycle-loop` — cycle loop mode (list → track → one → list)

## Changes

- **lib**: Added `refresh_all_feeds()` function to `lib/src/podcast/mod.rs`
- **server**: Restructured CLI with `PodcastAction` subcommand, wired `Refresh` in `execute_action()`
- **tui**: Restructured CLI with `PodcastAction` and `PlaylistAction` subcommands, wired both in `execute_action()` + added `execute_playlist_action()` via gRPC

## Notes

- Playlist commands (shuffle, sort, cycle-loop) require a running server — they connect via gRPC
- Podcast commands (import, export, refresh) run locally without a server
- This PR does NOT include the Invidious search fix (separate work)

re #304

### DIFF
--- a/lib/src/podcast/mod.rs
+++ b/lib/src/podcast/mod.rs
@@ -84,6 +84,22 @@ pub enum PodcastSyncResult {
     Error(PodcastFeed),
 }
 
+/// Dispatches `check_feed` for each podcast in the provided list.
+///
+/// This is the shared core used by both the CLI `refresh_all_feeds` and the TUI's
+/// `podcast_refresh_feeds`. Each caller provides its own callback to handle sync results.
+pub fn dispatch_feed_checks(
+    feeds: Vec<PodcastFeed>,
+    max_retries: usize,
+    taskpool: &TaskPool,
+    callback: impl Fn(PodcastSyncResult) + Send + 'static + Clone,
+) {
+    for feed in feeds {
+        let cb = callback.clone();
+        check_feed(feed, max_retries, taskpool, cb);
+    }
+}
+
 /// Spawns a new task to check a feed and retrieve podcast data.
 ///
 /// If `tx_to_main` is closed, no errors will be throws and the task will continue
@@ -259,6 +275,82 @@ fn duration_to_int(duration: Option<&str>) -> Option<i32> {
         1 => times[0],
         _ => None,
     }
+}
+
+/// Refreshes all podcast feeds in the database by re-fetching their RSS data
+/// and updating existing entries with new episodes.
+pub async fn refresh_all_feeds(db_path: &Path, config: &PodcastSettings) -> Result<()> {
+    let db_inst = Database::new(db_path)?;
+    let podcasts = db_inst.get_podcasts()?;
+    let total = podcasts.len();
+
+    if total == 0 {
+        println!("No podcasts to refresh.");
+        return Ok(());
+    }
+
+    println!("Refreshing {total} podcasts...");
+
+    let taskpool = TaskPool::new(usize::from(config.concurrent_downloads_max.get()));
+    let feeds: Vec<PodcastFeed> = podcasts
+        .into_iter()
+        .map(|pod| PodcastFeed::new(Some(pod.id), pod.url.clone(), Some(pod.title.clone())))
+        .collect();
+
+    let (tx_to_main, mut rx_to_main) = unbounded_channel();
+
+    dispatch_feed_checks(
+        feeds,
+        usize::from(config.max_download_retries),
+        &taskpool,
+        move |msg| {
+            let _ = tx_to_main.send(msg);
+        },
+    );
+
+    let mut failure = false;
+    while let Some(message) = rx_to_main.recv().await {
+        match message {
+            PodcastSyncResult::FetchPodcastStart(_) => (),
+            PodcastSyncResult::SyncData((id, pod)) => {
+                let title = &pod.title;
+                match db_inst.update_podcast(id, &pod) {
+                    Ok(_) => {
+                        println!("Updated {title}");
+                    }
+                    Err(err) => {
+                        failure = true;
+                        error!("Error updating {title}, err: {err}");
+                    }
+                }
+            }
+            PodcastSyncResult::NewData(pod) => {
+                let title = &pod.title;
+                // new data for an existing podcast shouldn't normally happen,
+                // but handle it gracefully by inserting
+                match db_inst.insert_podcast(&pod) {
+                    Ok(_) => {
+                        println!("Added {title}");
+                    }
+                    Err(err) => {
+                        failure = true;
+                        error!("Error adding {title}, err: {err}");
+                    }
+                }
+            }
+            PodcastSyncResult::Error(feed) => {
+                failure = true;
+                error!("Error retrieving RSS feed: {}", feed.url);
+            }
+        }
+    }
+
+    if failure {
+        bail!("Process finished with errors.");
+    }
+    println!("Refresh successful.");
+
+    Ok(())
 }
 
 /// Imports a list of podcasts from OPML format, reading from a file. If the `replace` flag is set, this replaces all
--- a/tui/src/cli.rs
+++ b/tui/src/cli.rs
@@ -1,6 +1,7 @@
 use clap::{ArgAction, Parser, Subcommand, ValueEnum, builder::ArgPredicate};
 use std::path::PathBuf;
 use termusiclib::config::v2::server::Backend as ConfigBackend;
+use termusiclib::player::{SortCriterion, SortDirection};
 
 #[derive(Parser, Debug)]
 // mostly read from `Cargo.toml`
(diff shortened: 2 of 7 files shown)


lib/src/podcast/mod.rs
```diff
@@ -261,6 +261,92 @@ fn duration_to_int(duration: Option<&str>) -> Option<i32> {
     }
 }
 
+/// Refreshes all podcast feeds in the database by re-fetching their RSS data
+/// and updating existing entries with new episodes.
+pub async fn refresh_all_feeds(db_path: &Path, config: &PodcastSettings) -> Result<()> {
```
hasezoey, line 282: Maybe it would be worth to de-duplicate this function with the one existing in the TUI at `tui/src/ui/components/podcast.rs` `Model::podcast_refresh_feeds`?

hasezoey, line 282: This function is quite CLI oriented, so it *should* live in a binary crate (`server` or `tui`), though as it is used by *both*, that is not possible without duplication.

We should likely consider only having it in one place or something.

lib/src/podcast/mod.rs
```diff
@@ -261,6 +261,92 @@ fn duration_to_int(duration: Option<&str>) -> Option<i32> {
...
+    }
+
+    println!("Refreshing {} podcasts...", podcasts.len());
+
+    let taskpool = TaskPool::new(usize::from(config.concurrent_downloads_max.get()));
+    let (tx_to_main, mut rx_to_main) = unbounded_channel();
+
+    for pod in &podcasts {
+        let tx_to_main_c = tx_to_main.clone();
+        let feed = PodcastFeed::new(Some(pod.id), pod.url.clone(), Some(pod.title.clone()));
```
hasezoey: Instead of taking by reference and then cloning, how about storing the total length before then (for later use) and consuming the array?

tui/src/cli.rs
```diff
@@ -106,14 +103,85 @@ pub enum Action {
...
+/// Sort criterion for playlist sorting.
+#[derive(Clone, Copy, Debug, ValueEnum)]
+pub enum SortCli {
+    Alphanumeric,
+    FirstAdded,
+    Duration,
+    MostPlayed,
+    Recency,
+    Frecency,
+}
```
hasezoey, line 162: What do you think about adding aliases like `LeastPlayed` / `LastAdded` / `HighestDuration` / `LowestDuration` practically act like `MostPlayed + DESC`, and if a `direction` is given, just invert it again?
Additionally for Alphanumeric, how about `AtoZ` and `ZtoA` (or something) alias?

Matrix for context:

|Input|Output Mode|Output Dir|
|:--:|:--:|:--:|
|FirstAdded (+ ASC)|FirstAdded|ASC|
|FirstAdded (+ DESC)|FirstAdded|DESC|
|LastAdded (+ ASC)|FirstAdded|DESC|
|LastAdded (+ DESC)|FirstAdded|ASC|

Alternatively, giving a direction with those extra aliases might also be set to be invalid, but if option `direction` is kept, why not just invert it.

hasezoey (changes requested):

tui/src/cli.rs
RG-IL: did it with a -i flag for invert

lib/src/podcast/mod.rs
RG-IL: fixed

lib/src/podcast/mod.rs
RG-IL: Shared fetch loop extracted into dispatch_feed_checks in lib. CLI and TUI each provide their own callback — CLI blocks and updates DB, TUI is event-driven. The orchestration differs, the core doesn't.

RG-IL: fixed

lib/src/podcast/mod.rs
```diff
@@ -261,6 +277,94 @@ fn duration_to_int(duration: Option<&str>) -> Option<i32> {
...
+        feeds,
+        usize::from(config.max_download_retries),
+        &taskpool,
+        move |msg| {
+            let _ = tx_for_dispatch.send(msg);
+        },
+    );
+
+    // drop the original sender so the channel closes when all tasks finish
+    drop(tx_to_main);
```
hasezoey: If you are not going to use the original instance anyway, why clone it?

lib/src/podcast/mod.rs
```diff
@@ -84,6 +84,22 @@ pub enum PodcastSyncResult {
...
+    feeds: Vec<PodcastFeed>,
+    max_retries: usize,
+    taskpool: &TaskPool,
+    callback: impl Fn(PodcastSyncResult) + Send + 'static + Clone,
+) {
+    for feed in feeds {
+        let cb = callback.clone();
+        check_feed(feed, max_retries, taskpool, cb);
+    }
+}
```
hasezoey, line 101: I didnt quite mean this way of de-duplication as this is rather small code, but we can let it stay.

tui/src/main.rs
```diff
@@ -532,33 +532,82 @@ fn get_path(dir: &Path) -> Result<PathBuf> {
...
+            playback.shuffle_playlist().await?;
+            println!("Shuffled playlist");
+        }
+        cli::PlaylistAction::Sort { criterion, invert } => {
+            let (criterion, should_invert) = criterion.into_criterion_and_invert();
+            let direction = if should_invert ^ invert {
+                termusiclib::player::SortDirection::Desc
+            } else {
+                termusiclib::player::SortDirection::Asc
+            };
```
hasezoey: Personally, i would prefer if the function converting the CLI `cli::PlaylistAction::Sort` would return `(SortCriterion, SortDirection)` and maybe `SortDirection` could just have a function `invert` which swaps / cycles through the values.

lib/src/podcast/mod.rs
hasezoey: Does not look that way.
From what i can tell, this PR does not touch `tui/src/ui/components/podcast.rs` at all.

hasezoey (changes requested):
Looks good and basically, done, just some minor styling.

Also please include `BREAKING CHANGE: <your reason>` in the commit message footer that modify outside breaking behavior (like moving `termusic export` to `termusic podcast export`, if there is no backwards compat alias). See [Commit messages: 4. For footer](https://github.com/tramhao/termusic/blob/master/CONTRIBUTING.md#commit-messages)

tui/src/cli.rs
```diff
@@ -106,14 +103,82 @@ pub enum Action {
...
+
+/// Playlist management subcommands.
+#[derive(Subcommand, Debug)]
+pub enum PlaylistAction {
+    /// Shuffle the current playlist.
+    Shuffle,
+    /// Sort the current playlist.
+    Sort {
+        /// Sort criterion.
+        #[arg(value_enum, default_value_t = SortCli::Alphanumeric)]
```
hasezoey: I just ran it to see how the ergonomics are.
I think it is counter intuitive to have a default on sort.

tui/src/cli.rs
```diff
@@ -106,14 +103,82 @@ pub enum Action {
...
+}
+
+/// Sort criterion for playlist sorting.
+#[derive(Clone, Copy, Debug, ValueEnum)]
+pub enum SortCli {
+    Alphanumeric,
+    #[value(alias = "atoz")]
+    AtoZ,
+    #[value(alias = "ztoa")]
+    ZtoA,
```
hasezoey:
```suggestion
    #[value(name = "atoz")]
    AtoZ,
    #[value(name = "ztoa")]
    ZtoA,
```

The default `kebab-case` rename makes those option, due to them being upper-lower-upper rename to `zto-a`, which is not great ergonomic wise and not intuitive, so we should manually *force* it to be `ztoa` instead of only having a alias (which in current clap are not written out).

lib/src/podcast/mod.rs
RG-IL: fixed

tui/src/cli.rs
RG-IL: fixed

tui/src/cli.rs
RG-IL: fixed :)

lib/src/podcast/mod.rs
```diff
@@ -261,6 +277,90 @@ fn duration_to_int(duration: Option<&str>) -> Option<i32> {
...
+            PodcastSyncResult::Error(feed) => {
+                msg_counter += 1;
+                failure = true;
+                error!("Error retrieving RSS feed: {}", feed.url);
+            }
+        }
+
+        if msg_counter >= total {
+            break;
+        }
```
hasezoey: Considering that the channel should be closed if all podcasts have been processed (which is the `while` condition), do we really need this?

hasezoey (approved):
One small question, otherwise looks done.

lib/src/podcast/mod.rs
RG-IL: You're right it probably shouldn't be there

PR #752, opened 2026-07-28
feat(tui): add custom loop mode display with Nerd Font support

### LINKED ISSUE #751: Allow custom symbols for loop/shuffle icons
## Describe what you want
Having a setting in tui.toml to put whatever symbol for loops icons, like what there is for the highlight symbols

Image

## Do you have already an idea for the implementation?
Somthing like this in the tui.toml

```toml
loop_symbol = "🔁"
shuffle_symbol = "🔀"
single_symbol = "🔂"
```

RG-IL:
Adds a `LoopModeDisplay` enum replacing the old `use_loop_mode_symbol` bool, supporting:

- `text` — plain text ("track", "playlist", etc.)
- `base_symbols` — emoji (🔂, 🔁, 🔀, ⮕) — default
- `nerd_font` — Nerd Font icons (󰑖, 󰑘, 󰒟, 󰒞)
- `custom` — user-defined symbols per mode

The old `use_loop_mode_symbol` field is deprecated but still read for backward compatibility. Config editor updated with a 3-option radio (Custom must be set manually in tui.toml).

fixes #751

### DIFF
--- a/lib/src/config/v2/tui/theme/styles.rs
+++ b/lib/src/config/v2/tui/theme/styles.rs
@@ -264,6 +264,7 @@ pub struct StylePlaylist {
 }
 
 impl StylePlaylist {
+    // TODO: consider adding a migration trait similar to what "Keys" has as "CheckConflict"
     #[must_use]
     pub fn effective_loop_mode_display(&self) -> LoopModeDisplay {
         match self.use_loop_mode_symbol_deprecated {
--- a/lib/src/config/v2/tui/theme/styles.rs
+++ b/lib/src/config/v2/tui/theme/styles.rs
@@ -23,29 +23,11 @@ pub enum LoopModeDisplay {
 }
 
 impl LoopModeDisplay {
+    /// Get the text to display for the given loopmode.
     #[must_use]
     pub fn display(&self, mode: LoopMode) -> &str {
         match self {
-            Self::Base(base) => match base {
-                LoopModeDisplayBase::Text => match mode {
-                    LoopMode::Track => "track",
-                    LoopMode::Playlist => "playlist",
-                    LoopMode::Random => "random",
-                    LoopMode::PlaylistOnce => "playlist once",
-                },
-                LoopModeDisplayBase::BaseSymbols => match mode {
-                    LoopMode::Track => "🔂",
-                    LoopMode::Playlist => "🔁",
-                    LoopMode::Random => "🔀",
-                    LoopMode::PlaylistOnce => "⮕",
-                },
-                LoopModeDisplayBase::NerdFont => match mode {
-                    LoopMode::Track => nf_loop_icons::TRACK,
-                    LoopMode::Playlist => nf_loop_icons::PLAYLIST,
-                    LoopMode::Random => nf_loop_icons::RANDOM,
-                    LoopMode::PlaylistOnce => nf_loop_icons::PLAYLIST_ONCE,
-                },
-            },
+            Self::Base(base) => base.display(mode),
             Self::Custom(symbols) => match mode {
                 LoopMode::Track => &symbols.track,
                 LoopMode::Playlist => &symbols.playlist,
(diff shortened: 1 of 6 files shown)


tui/src/ui/components/config_editor/general.rs
```diff
@@ -313,11 +313,25 @@ pub struct PlaylistDisplaySymbol {
...
+            termusiclib::config::v2::tui::theme::styles::LoopModeDisplay::Base(
+                termusiclib::config::v2::tui::theme::styles::LoopModeDisplayBase::BaseSymbols,
+            ) => 0,
+            termusiclib::config::v2::tui::theme::styles::LoopModeDisplay::Base(
+                termusiclib::config::v2::tui::theme::styles::LoopModeDisplayBase::NerdFont,
+            ) => 1,
+            _ => 2,
+        };
+        let component = common_radio_comp(&config_r, " Loop mode display mode? ")
+            .choices(["Base Symbols", "Nerd Font", "Text"])
```
hasezoey: This should have a option for `Custom`, but as it cannot be configured more it should also somehow warn when switching away / to it somehow (maybe via color?).
Otherwise, with the current implementation, you will just see option `Text` be selected when `Custom` is actually saved, which can be confusing.

tui/src/ui/components/config_editor/general.rs
```diff
@@ -313,11 +313,25 @@ pub struct PlaylistDisplaySymbol {
...
+            .theme
+            .style
+            .playlist
+            .effective_loop_mode_display();
+        let value = match &display {
+            termusiclib::config::v2::tui::theme::styles::LoopModeDisplay::Base(
+                termusiclib::config::v2::tui::theme::styles::LoopModeDisplayBase::BaseSymbols,
+            ) => 0,
+            termusiclib::config::v2::tui::theme::styles::LoopModeDisplay::Base(
+                termusiclib::config::v2::tui::theme::styles::LoopModeDisplayBase::NerdFont,
```
hasezoey: Please just import the types (even if just locally) to not have these long chains.

lib/src/config/v2/tui/theme/styles.rs
```diff
@@ -153,10 +230,23 @@ pub struct StylePlaylist {
...
+
+impl StylePlaylist {
+    #[must_use]
+    pub fn effective_loop_mode_display(&self) -> LoopModeDisplay {
+        match self.use_loop_mode_symbol_deprecated {
+            Some(true) => LoopModeDisplay::Base(LoopModeDisplayBase::BaseSymbols),
+            Some(false) => LoopModeDisplay::Base(LoopModeDisplayBase::Text),
+            None => self.loop_mode_display.clone(),
+        }
+    }
```
hasezoey, line 275: Personally, i would prefer if the old values get migrated right as the config gets read, instead of doing it at some point later.
This could be achieved via serde by doing a "Shadow" (Data-only) copy which is saved and the other is the runtime actual use value, via `#[serde(from = "DATAShadow")]`, see [this discussion](https://users.rust-lang.org/t/serde-populating-struct-fields-post-deserialize/114847).
Though that adds duplication, so another alternative is to have a function (similar to `CheckKeys` we already have) that runs once after the config has been read to migrate the data.

Though as this is the TUI config which practically only gets saved due to the config editor, the current implementation also works.

lib/src/config/v2/tui/theme/styles.rs
```diff
@@ -1,7 +1,84 @@
...
+}
+
+/// User-defined symbols for each loop mode.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct CustomLoopSymbols {
+    pub track: String,
+    pub playlist: String,
+    pub random: String,
+    pub playlist_once: String,
+}
```
hasezoey: Minor nitpick, please swap those 2 types location.

hasezoey (changes requested):

tui/src/ui/components/playlist/playlist_comp.rs
hasezoey, line 1: Due to the current implementation of `effective_loop_mode_display` and the config editor *not* clearing the deprecated value (setting to `None`), a change in the config editor does not apply until the TUI is restarted, which clears the value as serde does not write it out again.

tui/src/ui/components/config_editor/view.rs
```diff
@@ -656,15 +657,22 @@ impl Model {
...
-                .playlist
-                .use_loop_mode_symbol = matches!(display_symbol, 0);
+            let display = match display_idx {
+                3 => config_tui
+                    .settings
+                    .theme
+                    .style
+                    .playlist
+                    .loop_mode_display
+                    .clone(),
```
hasezoey: I dont like how that just clones the existing value. This should set it to `LoopModeDisplay::Custom`, if it is not already.
As for default values, some of the base ones can be cloned, i dont know which are the best option though.

lib/src/config/v2/tui/theme/styles.rs
```diff
@@ -1,7 +1,84 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum LoopModeDisplay {
+    Base(LoopModeDisplayBase),
+    Custom(CustomLoopSymbols),
+}
+
+/// User-defined symbols for each loop mode.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct CustomLoopSymbols {
```
hasezoey, line 50:
```suggestion
#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
#[serde(default)]
pub struct CustomLoopSymbols {
```

Otherwise, if the config does not contain *all* values, a error gets thrown. (more options *may* be added in the future)
Of course this requires a default impl though.

hasezoey (changes requested):

RG-IL:
As I cannot work on this until tomorrow afternoon, I will convert it into a draft temporarily.

tui/src/ui/components/playlist/playlist_comp.rs
RG-IL: You're right, my dad.

lib/src/config/v2/tui/theme/styles.rs
RG-IL: fixed

hasezoey (approved):
Looks good to me now. I made some minor style changes, are you fine with them?

RG-IL (commented):
Looks good to me

PR #760, opened 2026-08-26
feat(lib): load lyric from .lrc file

Yan710010:
When reading lyric from music metadata failed, and there's a `.lrc` file with the same name as the track file, it will read lyric from the .lrc file instead.

#### Two point I concerned

- I use `std::fs::read_to_string` and I'm not sure whether it would block.
- The original function seems able to return something like `Ok( Some( LyricData: { raw_lyrics: [], parsed_lyric: None } ) )`, in my patch I returned an error when both metadata and lrc file have no lyric, although this seems not to affect lyric displaying.

I'd like to improve it if needed.

### DIFF
--- a/lib/src/track/full_track.rs
+++ b/lib/src/track/full_track.rs
@@ -475,15 +475,34 @@ impl Track {
 
         let res = LYRIC_CACHE.with_borrow_mut(|cache| {
             cache
-                .try_get_or_insert(path_key, || {
-                    let result = parse_metadata_from_file(
+                .try_get_or_insert(path_key.clone(), || {
+                    let lyric_frames = parse_metadata_from_file(
                         track_data.path(),
                         MetadataOptions {
                             lyrics: true,
                             ..Default::default()
                         },
-                    )?;
-                    let lyric_frames = result.lyric_frames.unwrap_or_default();
+                    )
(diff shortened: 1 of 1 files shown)


hasezoey (commented):
This looks good, though i will need to test this more in-depth.

> I use std::fs::read_to_string and I'm not sure whether it would block.

Well, `parse_metadata_from_file` above is literally also just a blocking read, so it shouldnt matter.

> The original function seems able to return something like Ok( Some( LyricData: { raw_lyrics: [], parsed_lyric: None } ) ), in my patch I returned an error when both metadata and lrc file have no lyric, although this seems not to affect lyric displaying.

I think the original intention behind the empty lyric data was because the `LRUCache` we use, at least its type, stores a lyric object, not a Option or Result, so in the no-error case & no lyric case, we wanted to store the result so it is not tried again.
I guess we should document what may get returned for future reference.

hasezoey (approved):
This looks to be working without issue for normal playback.

It likely should also be possible to load a LRC file for the tag editor, though that can be a follow-up PR.

Yan710010:
> so in the no-error case & no lyric case, we wanted to store the result so it is not tried again.

That sounds reasonable, I'll fix it.

PR #761, opened 2026-08-26
fix(lib::track::full_track): return empty lyric data instead of error

Yan710010:
> [#760](https://github.com/tramhao/termusic/pull/760#pullrequestreview-5030119226)
... so in the no-error case & no lyric case, we wanted to store the result so it is not tried again.

---

Since I really confused on the type `try_get_or_insert` returned: `Result<_, Option<Error>>` (I cannot understand why theres an Option)...

> I guess we should document what may get returned for future reference.

For now I use `.map_err(|e| Some(e.into()))` to be compatible with that type while I'd like to take some change if needed.

### DIFF
--- a/lib/src/songtag/lrc.rs
+++ b/lib/src/songtag/lrc.rs
@@ -1,19 +1,22 @@
-// Package downloaded lyrics from different websites and embed them into an MP3 file.
-// lrc file is used to parse lrc file into subtitle. Similar to subtitles package
-// [al:''Album where the song is from'']
-// [ar:''Lyrics artist'']
-// [by:''Creator of the LRC file'']
-// [offset:''+/- Overall timestamp adjustment in milliseconds, + shifts time up, - shifts down'']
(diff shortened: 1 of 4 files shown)


hasezoey:
> Since I really confused on the type try_get_or_insert returned: Result<_, Option<Error>> (I cannot understand why theres an Option)...

I was also now stumped on this, and i think it was because i just copied it from the `PICTURE_CACHE`, where it did make use of `Err(Option)` to signal that there was no picture present, but not store so in the cache.

---

I have made some modifications to clarify the situation for now and the future.
It also contains a bug-fix for embedded lyrics not being read (since likely the upgrade to lofty 0.24).
Are you fine with the changes, or have any extra comments?

EDIT: force-pushed to fix a quick typo.

Yan710010 (commented):
I think it's fine

hasezoey (approved):

PR #768, opened 2026-08-30
fix(tui::ui): add change theme msg to reload theme

### LINKED ISSUE #722: After applying a new theme, some components still have old theme applied
## Environment & Versions

- System: NixOS unstable
- Termusic version: 2cdd3e095cda089482ddf5448bc55d462b945d93
- Rust version: 1.95.0

Logs:

logs?

## Description of the Problem

If I change theme, some of components doesn't apply it.

Before selected theme:

Image

Selecting theme:

I suppose the top tab list & save changes should update theme too?

Image

After selecting theme:

Library still use old theme
Image

This one is normal
Image

Podcast & Episodes aren't updated
Image

Also the top right song view when doesn't change theme. Sorry for bad image my notifications seem cover it but you can see some of them image still have white old theme
.

theeasternfurry:
Fixes #722

## What I did (Edit)
- Add `Msg::ChangeTheme` and dispatch a synthetic event to the affected  static components so they refresh their colors after a theme change.
- Add new methods & doc-comment for `apply_theme_style` and `refresh_theme`

### DIFF
--- a/lib/src/config/v2/tui/theme/mod.rs
+++ b/lib/src/config/v2/tui/theme/mod.rs
@@ -427,6 +427,9 @@ impl ThemeColors {
             bright: ThemeBright::native(),
         }
     }
+
+    /// How many Themes there are without actual files and always exist
+    pub const THEMES_WITHOUT_FILES: usize = 2;
 }
 
 /// Error for when [`ThemeColors`] parsing fails
--- a/lib/src/config/v2/tui/theme/mod.rs
+++ b/lib/src/config/v2/tui/theme/mod.rs
@@ -427,9 +427,6 @@ impl ThemeColors {
             bright: ThemeBright::native(),
         }
     }
-
-    /// How many Themes there are without actual files and always exist
-    pub const THEMES_WITHOUT_FILES: usize = 2;
 }
 
 /// Error for when [`ThemeColors`] parsing fails
--- a/tui/src/ui/components/config_editor/update.rs
+++ b/tui/src/ui/components/config_editor/update.rs
@@ -5,7 +5,6 @@ use termusiclib::config::v2::tui::config_extra::TuiConfigVersionedDefaulted;
 use termusiclib::config::v2::tui::keys::KeyBinding;
 use termusiclib::config::v2::tui::theme::ThemeColors;
 use termusiclib::config::v2::tui::theme::styles::ColorTermusic;
-use termusiclib::utils::get_app_config_path;
 
 use crate::ui::Model;
 use crate::ui::components::CEHeader;
@@ -71,6 +70,8 @@ impl Model {
                         if both_ok {
                             self.command(TuiCmd::ReloadConfig);
 
+                            self.refresh_static_theme_components();
+
                             // only exit config editor if saving was successful
                             self.umount_config_editor();
                         }
@@ -126,40 +127,10 @@ impl Model {
 
     /// Preview theme at Table index
     fn preview_theme(&mut self, index: usize) {
-        // table entry 0 is termusic default
-        if index == 0 {
-            self.preview_theme_apply(ThemeColors::full_default(), 0);
-
+        let Some(theme) = self.resolve_theme_by_index(index) else {
             return;
-        }
-        if index == 1 {
-            self.preview_theme_apply(ThemeColors::full_native(), 1);
-
-            return;
-        }
-
-        // idx - THEMES_WITHOUT_FILES as 0 until THEMES_WITHOUT_FILES table-entries are termusic themes without files, which always exists
-        if let Some(theme_filename) = self.config_editor.themes.get(index - THEMES_WITHOUT_FILES) {
-            match get_app_config_path() {
-                Ok(mut theme_path) => {
-                    theme_path.push("themes");
-                    theme_path.push(format!("{theme_filename}.yml"));
-                    match ThemeColors::from_yaml_file(&theme_path) {
-                        Ok(mut theme) => {
-                            theme.file_name = Some(theme_filename.clone());
-
-                            self.preview_theme_apply(theme, index);
-                        }
-                        Err(e) => {
-                            error!("Failed to load theme colors: {e:?}");
-                        }
-                    }
-                }
-                Err(e) => {
-                    error!("Error getting config path: {e:?}");
-                }
-            }
-        }
+        };
+        self.preview_theme_apply(theme, index);
     }
 
     /// Apply the given theme as a preview
--- a/tui/src/ui/components/config_editor/update.rs
+++ b/tui/src/ui/components/config_editor/update.rs
@@ -5,16 +5,22 @@ use termusiclib::config::v2::tui::config_extra::TuiConfigVersionedDefaulted;
 use termusiclib::config::v2::tui::keys::KeyBinding;
 use termusiclib::config::v2::tui::theme::ThemeColors;
 use termusiclib::config::v2::tui::theme::styles::ColorTermusic;
+use termusiclib::utils::get_app_config_path;
+use tuirealm::event::Event;
 
 use crate::ui::Model;
 use crate::ui::components::CEHeader;
 use crate::ui::ids::{Id, IdCETheme, IdConfigEditor, IdKey, IdKeyGlobal, IdKeyOther};
+use crate::ui::model::UserEvent;
 use crate::ui::msg::{
     CONFIG_EDITOR_TABS_ORDER, ConfigEditorLayout, ConfigEditorMsg, GENERAL_FOCUS_ORDER,
-    KFGLOBAL_FOCUS_ORDER, KFMsg, KFOTHER_FOCUS_ORDER, THEME_COLOR_ITEM_FOCUS_ORDER_START,
+    KFGLOBAL_FOCUS_ORDER, KFMsg, KFOTHER_FOCUS_ORDER, Msg, THEME_COLOR_ITEM_FOCUS_ORDER_START,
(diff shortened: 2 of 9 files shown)


lib/src/config/v2/tui/theme/mod.rs
```diff
@@ -427,6 +427,9 @@ impl ThemeColors {
             bright: ThemeBright::native(),
         }
     }
+
+    /// How many Themes there are without actual files and always exist
+    pub const THEMES_WITHOUT_FILES: usize = 2;
```
hasezoey: Personally, i dont quite like having this here, as the lib does not really use it anywhere, only the TUI component actually knows what static themes are present.

tui/src/ui/model/update.rs
```diff
@@ -1268,4 +1274,71 @@ impl Model {
...
+        let ev = Event::User(UserEvent::Forward(Msg::ChangeTheme(0)));
+
+        for id in [Id::Library, Id::Podcast, Id::Episode, Id::MessagePopup] {
+            if let Some(component) = self.app.get_component_mut(&id) {
+                component.on(&ev);
+            }
+        }
+    }
+
+    pub(crate) fn resolve_theme_by_index(&self, index: usize) -> Option<ThemeColors> {
```
hasezoey: Even though this is internal and the name is self-descriptive, it would be good to have some doc-comment. (ex. mention that `0` and `1` resolve to special themes)

tui/src/ui/msg.rs
```diff
@@ -51,6 +51,8 @@ pub enum Msg {
     /// Note that this message does *not* update the drawn cover.
     ForceRedraw,
 
+    ChangeTheme(usize),
```
hasezoey: What does the `usize` argument do? From what i can tell, currently it is hardcoded to `0`.

Also, personally i would prefer the name `ReloadTheme`, WDYT?

tui/src/ui/model/update.rs
```diff
@@ -1268,4 +1274,71 @@ impl Model {
...
+    /// Re-apply theme colors to components that bake style in at mount time
+    /// and don't otherwise re-read `config_tui` on every render (unlike e.g. Playlist).
+    pub fn refresh_static_theme_components(&mut self) {
+        let ev = Event::User(UserEvent::Forward(Msg::ChangeTheme(0)));
+
+        for id in [Id::Library, Id::Podcast, Id::Episode, Id::MessagePopup] {
+            if let Some(component) = self.app.get_component_mut(&id) {
+                component.on(&ev);
+            }
+        }
```
hasezoey: While this likely works, this completely circumvents the existing event subscriptions propagation, which i would expect to find this in instead of some other file.

tui/src/ui/model/update.rs
```diff
@@ -1268,4 +1274,71 @@ impl Model {
...
+
+        self.refresh_static_theme_components();
+
+        let save_result = {
+            let config = self.config_tui.read();
+            TuiConfigVersionedDefaulted::save_config_path(&config.settings)
+        };
+        if let Err(e) = save_result {
+            self.mount_error_popup(e.context("save theme"));
+        }
```
hasezoey: Unless i missed something, this practically unncessary as [the config editor saves the config's new theme already](https://github.com/theeasternfurry/termusic/blob/e6ab7d8ee0fe0821ba1290aee476b2f2ea9b195d/tui/src/ui/components/config_editor/view.rs#L636). (aside from `refresh_static_theme_components`, but github does not let me exclude that line)

The only time i see this being useful is if somehow outside of the config editor the theme needs to be changed, but such action does not currently exist, nor do i think it will in the (near) future.
Even if it does in the future, the config writing should likely be done before sending out the event (ex. in a debug settings explorer).

tui/src/ui/components/orx_music_library/music_library.rs
```diff
@@ -736,6 +736,30 @@ impl OrxMusicLibraryComponent {
...
+            std::mem::swap(self.component.get_tree_mut(), fresh.get_tree_mut());
+        }
+
+        self.component = fresh;
+
+        if let Some(idx) = selected_idx {
+            self.component.select(idx);
+            self.component.open_all_parents(idx);
+            self.component.open(idx);
+        }
```
hasezoey: Instead of creating a new tree and copying over the actual state, why not just modify the theme options?

Functions like `foreground` are literally just easy helpers for calling `comp.attr(ATTR, VALUE)`, see [this code](https://docs.rs/tuirealm-orx-tree/0.4.0/src/tuirealm_orx_tree/component.rs.html#146-150). (meaning `comp.attr` could be called manually)

Alternatively, you could just re-use those self-taking functions:
```rust
let moved = std::mem::take(&mut self.component); // or ::replace(&mut self.component, DEFAULT) if automatic Default cannot be used due to generic, see default creation in get_inner_comp

let moved = moved.forgeround(VALUE).background(VALUE) // etc..

let _ = std::mem::replace(&mut self.component, moved);
```
And/or additionally modify `get_inner_comp` to take a `TreeView` instance to apply those values to instead of creating one in the function itself. (or just extract the theme related options into another function).

hasezoey (changes requested):
Thanks for taking your time to resolve this issue.

This looks like a good initial approach, but there are some better ways to avoid duplication and usage of `unsafe`. (see comments)

Having to copy state is the thing i wanted to avoid, as we may not know what the components actual state is. (for example many new components [store the scroll offset](https://github.com/hasezoey/tuirealm-orx-tree/blob/33e5484cc44b2a394c403ce867fe1222f1bc8c38/src/state.rs#L91-L94) instead of statically being X amount after the selected offset, which would be lost in this case)

tui/src/ui/msg.rs
theeasternfurry: You're right, it's not used at all here — I think I initially assumed I'd need the index, but turns out I didn't in the end.

`ReloadTheme` fits this better than reusing `ChangeTheme`. I'll split it out.

lib/src/config/v2/tui/theme/mod.rs
theeasternfurry: I am not sure if I want to put it back in `tui/src/ui`.  I also have problem with `resolve_theme_by_index` since it also need `THEMES_WITHOUT_FILES`

You have any other suggestion?

tui/src/ui/model/update.rs
theeasternfurry: I will add old doc-comment.

I migrate it from config_editor to this but `pub(crate)` make this one ugly and config_editor recall this method. Do you have problem with it? Should I move it else where with `THEMES_WITHOUT_COLORS` since it only return `ThemeColors` directly? Somewhere in `lib/src/config/v2/tui/theme/`?

lib/src/config/v2/tui/theme/mod.rs
hasezoey: My reasoning is that `THEMES_WITHOUT_FILES` is specific to the TUI config theme editor as only it knows what themes it actually displays where. Hence why it should be there. (though i guess even previously, it was in a different spot)

tui/src/ui/model/update.rs
hasezoey: > I migrate it from config_editor to this but pub(crate) make this one ugly and config_editor recall this method. Do you have problem with it? Should I move it else where with THEMES_WITHOUT_COLORS since it only return ThemeColors directly? Somewhere in lib/src/config/v2/tui/theme/?

Yes, as it directly makes use of `THEMES_WITHOUT_COLORS` and with the other changes, it is practically only used by the TUI config theme editor, it should likely also be there.

tui/src/ui/components/config_editor/update.rs
```diff
@@ -124,6 +127,18 @@ impl Model {
...
+    /// Re-apply theme colors to components that bake style in at mount time
+    /// and don't otherwise re-read `config_tui` on every render (unlike e.g. Playlist).
+    fn refresh_static_theme_components(&mut self) {
+        let ev = Event::User(UserEvent::Forward(Msg::ReloadTheme));
+
+        for id in [Id::Library, Id::Podcast, Id::Episode, Id::MessagePopup] {
+            if let Some(component) = self.app.get_component_mut(&id) {
+                component.on(&ev);
+            }
+        }
```
hasezoey, line 139: I still think this should be changed to be handled via subscriptions.
Though that can also be done in a follow-up, or i can push a commit to do it.

tui/src/ui/components/popups/message.rs
hasezoey, line 1: To not accidentally forget to change one place, this likely should make use of the same pattern as the `music_library` does for reloading (by taking).

tui/src/ui/components/podcast.rs
hasezoey, line 1: Once again here, this should not create a new component if it can avoid it.

Also doc-comments are missing.

hasezoey (changes requested):

tui/src/ui/components/config_editor/update.rs
theeasternfurry: @hasezoey I haven't worked with the subscription API before, so I'd rather leave this one to you to avoid more churn on an already PR. Happy to pull it and test it manually once pushed, or take it on separately later once I've looked into it.

hasezoey (approved):
Code looks great and confirmed it works.

No need to fix the typo in the commit message, i am gonna squash merge this.

tui/src/ui/components/config_editor/update.rs
hasezoey: See #770

PR #770, opened 2026-08-31
Use the subscription API for reloading theme

hasezoey:
Re https://github.com/tramhao/termusic/pull/768#discussion_r3889923362

#768 added a way to reload the theme, but did not make use of the subscription API. This PR changes that to have everything consistent and automatically populate out instead of changing a static list.

Tagging @theeasternfurry as you might be curious.


theeasternfurry:
It works. Say only to confirm it again.

PR #773, opened 2026-09-01
Split the massive proto file into single modules

hasezoey:
I finally got it working, i had tried this last year, but my IDE's (vscode) extension did not like it, making it harder to work with.
But now everything "just works", though with the trade-off that we have to manually define each new proto file in the `build.rs` for now.

The massive proto and single service is now split into:
- server (everything that does not fit anywhere else)
- player (directly player control)
- queue (playlist manipulation)
- stream (stream update related things)

This split does not just apply to the proto files (as can be seen by the massive change list), but also the client and server files as well.

Additionally, also adds a `editorconfig` file, as otherwise my IDE had indentation all over the place in `proto` files for some reason.


hasezoey:
I thought a bit, and i am gonna clean-up the commit history a bit tomorrow, instead of having one massive one.

tramhao:
Very nice job done:)

PR #778, opened 2026-09-11
fix: serialize concurrent library database migrations

### LINKED ISSUE #767: [Bug]: Concurrent database migration race condition on fresh startup ("duplicate column name: total_play_count")
It could totally be me, but here goes anyway. Happened to me after fresh/first install. I deleted both `data.db` and `library2.db` files and `termusic` started fine after.

### Describe the bug

On a clean install / fresh launch (when no `~/.config/termusic/` database exists), launching `termusic` spawns `termusic-server` in the background. Both the TUI and server processes concurrently call database initialization (`migrate(conn)`).

Because migrations are not locked in an exclusive transaction across steps, both processes race during the migration from version 0 -> 1 -> 2:
1. One process executes `001.sql` and sets `PRAGMA user_version = 1`.
2. The other process sees `user_version == 1` and tries to execute `002.sql` (`ALTER TABLE tracks ADD COLUMN total_play_count ...`).
3. Since both processes run `002.sql` concurrently, SQLite errors out with `duplicate column name: total_play_count`, causing a panic.

### Log / Backtrace

```text
thread 'main' (11796) panicked at tui/src/ui/model/mod.rs:397:47:
Open Library Database: Database migration

Caused by:
    0: Database version 2 migration failed
    1: duplicate column name: total_play_count
    2: Error code 1: SQL logic error
```

### Steps to Reproduce

1. Remove any existing database (`rm -rf ~/.config/termusic/*.db`).
2. Run `termusic` on a clean environment.
3. Observe intermittent panic on startup during migration.

### Possible Solution

In `lib/src/new_database/migrate.rs`:
- Wrap the entire migration process in an exclusive transaction (`BEGIN EXCLUSIVE` / `conn.exclusive_transaction()`), or
- Set `PRAGMA busy_timeout` on the SQLite connection, or
- Let `termusic-server` handle database migrations and have the TUI connect once the server reports ready.

mvanhorn:
Change `migrate` to accept `&mut Connection` and acquire a `transaction_with_behavior(TransactionBehavior::Immediate)` before the first schema-version read, ensuring competing initializers wait before deciding which migrations remain. On a fresh installation, the TUI and background server can initialize the same library database concurrently and panic with `duplicate column name: total_play_count` during migration 002.

Fresh creation: retain the table assertions and additionally verify version 2, one instance of each added column, and all three metadata keys. Re-running migration must preserve schema, data, creation metadata, and `last_migrated_at` (seed a sentinel value to make this assertion independent of clock precision); Existing version 1: initialize from the actual 001 SQL and set `user_version = 1`, insert a representative track with a known RFC3339 `added_at`, then migrate. Verify preserved row identity, converted integer timestamp, `total_play_count = 0`, and null `last_played_at`.

Fixes #767


lib/src/new_database/migrate.rs
```diff
@@ -25,8 +25,13 @@ fn set_user_version(conn: &Connection, version: u32) -> Result<u32> {
 }
 
 /// Check and update the database to be at [`DB_VERSION`].
-pub(super) fn migrate(conn: &Connection) -> Result<()> {
-    let user_version: u32 = get_user_version(conn)?;
+pub(super) fn migrate(conn: &mut Connection) -> Result<()> {
+    // Reserve the writer before reading the version so competing initializers
+    // decide which migrations remain only after the previous writer commits.
+    let tx = conn
+        .transaction_with_behavior(TransactionBehavior::Immediate)
```
hasezoey, line 32: I have not used this before, but wouldnt `Exclusive` be more fitting here?

lib/src/new_database/migrate.rs
```diff
@@ -113,21 +119,148 @@ fn set_db_created_with(conn: &Connection) -> Result<()> {
...
+            if !thread::panicking() {
+                result.unwrap();
+            }
+        }
+    }
+
+    fn version_one(conn: &Connection) {
+        conn.execute_batch(include_str!("./migrations/001.sql"))
+            .unwrap();
+        set_user_version(conn, 1).unwrap();
```
hasezoey, line 175: If we need specific versions, how about splitting `apply_migrations`'s `if` blocks to their own functions?

lib/src/new_database/migrate.rs
```diff
@@ -113,21 +119,148 @@ fn set_db_created_with(conn: &Connection) -> Result<()> {
...
+        conn.execute(
+            "INSERT INTO tracks(id, file_dir, file_stem, file_ext, added_at)
+             VALUES (42, '/music', 'example', 'mp3', ?1)",
+            [ADDED_AT],
+        )
+        .unwrap();
+    }
+
+    fn schema(conn: &Connection) -> Vec<(String, String)> {
```
hasezoey, line 184: Personally, i would rename that to `get_all_schemas`

lib/src/new_database/migrate.rs
```diff
@@ -113,21 +119,148 @@ fn set_db_created_with(conn: &Connection) -> Result<()> {
...
+             WHERE name NOT LIKE 'sqlite_%' ORDER BY name",
+        )
+        .unwrap()
+        .query_map([], |r| Ok((r.get(0)?, r.get(1)?)))
+        .unwrap()
+        .collect::<rusqlite::Result<_>>()
+        .unwrap()
+    }
+
+    fn metadata(conn: &Connection) -> Vec<(String, String)> {
```
hasezoey, line 196: Personally, i would rename that to `get_all_config_data`, or something along those lines.

lib/src/new_database/migrate.rs
```diff
@@ -113,21 +119,148 @@ fn set_db_created_with(conn: &Connection) -> Result<()> {
...
+        conn.prepare("SELECT key, value FROM config ORDER BY key")
+            .unwrap()
+            .query_map([], |r| Ok((r.get(0)?, r.get(1)?)))
+            .unwrap()
+            .collect::<rusqlite::Result<_>>()
+            .unwrap()
+    }
+
+    fn assert_creation_metadata(conn: &Connection) {
```
hasezoey, line 205:
```suggestion
    /// Asserts that all config key-value pairs from the migration exist and are correct for DB version 2.
    fn assert_creation_metadata(conn: &Connection) {
```

lib/src/new_database/migrate.rs
```diff
@@ -113,21 +119,148 @@ fn set_db_created_with(conn: &Connection) -> Result<()> {
...
+        let values = metadata(conn);
+        assert_eq!(values.len(), 3);
+        assert_eq!(values[0].0, "db_created_at");
+        chrono::DateTime::parse_from_rfc3339(&values[0].1).unwrap();
+        assert_eq!(values[1], ("db_created_with".into(), crate::VERSION.into()));
+        assert_eq!(values[2].0, "last_migrated_at");
+        chrono::DateTime::parse_from_rfc3339(&values[2].1).unwrap();
+    }
+
+    fn assert_migrated_track(conn: &Connection) {
```
hasezoey, line 215:
```suggestion
    /// Assert that the track from [`version_one`] is correctly migrated to version 2.
    fn assert_migrated_track(conn: &Connection) {
```

lib/src/new_database/migrate.rs
```diff
@@ -157,5 +290,213 @@ mod tests {
...
+            .query_map([], |r| Ok((r.get(0)?, r.get(1)?)))
+            .unwrap()
+            .collect::<rusqlite::Result<_>>()
+            .unwrap();
+        for name in ["total_play_count", "last_played_at", "added_at"] {
+            assert_eq!(columns.iter().filter(|(col, _)| col == name).count(), 1);
+            assert!(columns.contains(&(name.into(), "INTEGER".into())));
+        }
+        assert!(!columns.iter().any(|(name, _)| name == "added_at_new"));
+    }
```
hasezoey, line 306: In the interest of keeping the functions together, this should be just after the `assert_migrated_track` function, before the test it is currently at.

hasezoey (changes requested):
Thank you for addressing this issue and improving tests. Mostly looks fine, just some minor style here and there.

mvanhorn:
Thanks for the review. All six points addressed in f2cd44c:

- `Exclusive` instead of `Immediate`: agreed, that is the better fit. The migration now takes a full write lock before reading `user_version`, so a competing writer waits on the transaction rather than racing the version read.
- Split `apply_migrations`' if-blocks: they are now `apply_version_1` and `apply_version_2`. The tests that needed a specific version (`version_one`, the competing-v2-apply case) call those helpers instead of duplicating the SQL.
- `schema` -> `get_all_schemas`.
- `metadata` -> `get_all_config_data`.
- Both suggested doc comments applied as written.
- `assert_full_schema` moved up next to the other assert helpers, right after `assert_migrated_track`.

Verified: `cargo test -p termusic-lib --lib new_database::migrate` 7/7 passing, `cargo clippy -p termusic-lib --all-targets -- -D warnings` clean, `cargo fmt` clean.

hasezoey (approved):
Thank you for fixing this issue.
I made a small change to make all the `apply_version_*` functions more consistent.

Next time, feel free to resolve/close the threads that you think are fully resolved.

